PR #98, opened 2016-02-19
Made worksheets work again, added a tutorial for worksheets

### LINKED ISSUE #95: Document worksheets, in particular AutoDocWorksheet
The whole worksheet business is currently undocumented. Please fix that.

In particular, I would like to refacto the code in `gap/Magic.gi`, to make it less of a hack, and make it clearer which bits are used for worksheets, which for packages, and which for both.

But I cannot really do that, as I don't know what is supposed to work and how; so if I find some weird edge behavior of code, I have no way of knowing whether it is intentional, or a bug, or perhaps a case of "used to be intentional but now we don't use it anymore".

As a minor first step, I just opened sebasguts/AutoDocTestPackage#5 which at least adds a single test worksheet (one that @sebasguts once sent me)

sebasguts:
Closes #95. More refactoring to come here.

Could you have a look whether this tutorial about worksheets helps to understand their functionality?

### DIFF
--- a/doc/Tutorials.xml
+++ b/doc/Tutorials.xml
@@ -264,4 +264,55 @@ function.
 
 </Section>
 
+<Section Label="Tut:AutoDocWorksheet">
+  <Heading>AutoDoc worksheets</Heading>
+&AutoDoc; worksheets can be used to create html and pdf documents using AutoDoc syntax and possibly
+including &GAP; examples and implementations without having them associated to a package.
+A file for a worksheet could look like this:
+<Listing>
+#! @Title My first worksheet
+#! @Author Charlie Brown
+
+#! @Chapter Some groups
+
+#! @BeginExample
+S3 := SymmetricGroup( 3 );;
+S4 := SymmetricGroup( 4 );;
+#! @EndExample
+</Listing>
+Now, one can create a pdf and html document, like a package documentation out of it. Suppose the document above
+is saved as <F>worksheet.g</F>. Then, when GAP is started in the folder of this file, the command
+<Listing>
+AutoDocWorksheet( "worksheet.g" );
+</Listing>
+will create a subfolder called <F>doc</F> of the current directory in which it will create the documentation.
+There are several options to configure the output of the worksheet command, which are identical to the
+options of the <Ref Func="AutoDoc"/> command. It is even possible to test the examples in the worksheet using the maketest option
+from the AutoDoc command.
+<P/>
+Since the worksheets do not have a <F>PackageInfo.g</F> to extract information, all possible tags that &GAPDoc; supports for
+the title page can be set into the document. A fully typed titlepage can look like this:
+<Listing>
+#! @Title My first worksheet
+#! @Subtitle Some small examples
+#! @Author Charlie Brown
+
+#! @Version 0.1
+#! @TitleComment Some worksheet
+#! @Date 01/01/2016
+#! @Address TU Kaiserslautern
+#! @Abstract
+#!  A worksheet showing some small examples about groups.
+#! @Copyright 2016 Charlie Brown
+#! @Acknowledgements Woodstock
+#! @Colophon Some colophon
+
+#! @Chapter Some groups
+
+#! @BeginExample
+S3 := SymmetricGroup( 3 );;
+S4 := SymmetricGroup( 4 );;
+#! @EndExample
+</Listing>
+</Section>
 </Chapter>
--- a/gap/AutoDocMainFunction.gd
+++ b/gap/AutoDocMainFunction.gd
@@ -44,9 +44,11 @@ DeclareGlobalFunction( "CreateMakeTest" );
 #! @Section Worksheets
 
 #! @Description
-#!  This function works exactly like the AutoDoc command, except that no package is needed to create a
-#!  worksheet. It takes the sampe optional records as the AutoDoc-command, so please refer this command
-#!  for a full list.
-#!  It's only optional argument is a (list of) filenames, which are then scanned by the AutoDoc parser.
+#!  The intention of these function is to create stand-alone pdf and html files
+#!  using AutoDoc without having them associated to a package.
+#!  It uses the same optional records as the &AutoDoc; command itself, but instead of
+#!  a package name there should be a filename or a list of filenames containing AutoDoc
+#!  text from which the documents are created. Please see the &AutoDoc; command for more
+#!  information about this and have a look at <Ref Label="Tut:AutoDocWorksheet"/> for a simple worksheet example.
 #! @Arguments list_of_filenames : options
 DeclareGlobalFunction( "AutoDocWorksheet" );
--- a/gap/AutoDocMainFunction.gi
+++ b/gap/AutoDocMainFunction.gi
@@ -304,7 +304,7 @@ end );
 
 ##
 ## Optional argument is PackageName, which creates a
-## Default chapter record. This is not availible for
+## Default chapter record. This is not available for
 ## worksheets.
 InstallGlobalFunction( AutoDocScanFiles,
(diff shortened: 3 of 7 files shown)


gap/AutoDocMainFunction.gi
```diff
@@ -304,7 +304,7 @@ end );
 
 ##
 ## Optional argument is PackageName, which creates a
-## Default chapter record. This is not availible for
+## Default chapter record. This is not availiable for
```
fingolfin: still wrong, try "available" :-)

gap/DocumentationTree.gi
```diff
@@ -324,6 +324,12 @@ InstallMethod( Add, [ IsTreeForDocumentationNode, IsString ],
 end );
 
 ##
+InstallMethod( Add, [ IsTreeForDocumentation, IsString ],
+  function( tree, string )
+    Add( tree!.content, string );
+end );
```
fingolfin: Please add a comment that explain when this method is used (I assume it is workspace specific, as I wasn't able to trigger it in my tests with packages).

gap/Parser.gi
```diff
@@ -585,27 +585,30 @@ InstallGlobalFunction( AutoDoc_Parser_ReadFiles,
...
-
-            Add( tree, current_command[ 2 ] );
+            current_item := tree!.TitlePage.( name );
+            Add( current_item, current_command[ 2 ] );
         end;
+    end;
+
+    for title_item in title_item_list do
+        command_function_record.( Concatenation( "@", title_item ) ) := create_title_item_function( title_item );
     od;
```
fingolfin: I see you are reverting parts of my previous edit here. You probably have a good reason, but to prevent us changing this back and forther, you should documented you reason in the form of a comment here.

gap/Parser.gi
```diff
@@ -585,27 +585,30 @@ InstallGlobalFunction( AutoDoc_Parser_ReadFiles,
...
             fi;
-            tree!.content := tree!.TitlePage.TitleComment;
+            current_item := tree!.TitlePage.TitleComment;
 
-            Add( tree, "<URL>" );
-            Add( tree, current_command[ 2 ] );
-            Add( tree, "</URL>" );
+            Add( current_item, "<URL>" );
+            Add( current_item, current_command[ 2 ] );
+            Add( current_item, "</URL>" );
```
fingolfin: So, I actually wonder about this @URL -- it is listed in the AutoDoc manual in section "2.3 Title page commands" together with all the other title page commands. But clearly it is different from the rest, since you have to implement it differently. And indeed, I just had a look at the GAPDoc manual, and there is no URL element of the title page.

So, section "2.3 Title page commands" should be adapted, documenting @URL differently.

Or perhaps @URL should be abolished -- I don't understand what it is good for, why can't you just write
```
#! @TitleComment <URL>http://bla</URL>
```
or even (if we automatically turned URLs into links)
```
#! @TitleComment http://bla
```
After all, is it really that common that one wants to put an URL into a TitleComment that it justifies having a dedicated AutoDoc command for it?

gap/Parser.gi
```diff
@@ -585,27 +585,30 @@ InstallGlobalFunction( AutoDoc_Parser_ReadFiles,
...
         end
     );
     title_item_list := [ "Title", "Subtitle", "Version", "TitleComment", "Author",
                          "Date", "Address", "Abstract", "Copyright", "Acknowledgements", "Colophon" ];
```
fingolfin: And while I am nitpicking about things only partially related to this pull request: I think it would be really good if the manual contained an example showing of how to use these commands.

gap/Parser.gi
fingolfin: To clarify: I am talking about 188c7e56dc539f6229ecedf7bf28b419c5203665 here. There, I got rid of `install_tmp_func`, which you now reintroduced as `create_title_item_function`.

fingolfin: Oh, wait, now I see: the problem is that the value of `title_item` keeps changing. So my previous commit was wrong, and this change is neccesary.

sebasguts: Ah, okay, I see. The problem, if you don't do it, is that the function added to the record will have the value of ```title_item``` stored to the last list entry, ```Colophon``` - this is how ```GAP``` behaves. Having the function around it solves that.

fingolfin: Yup. I should have known, I've run into this trap before :-). So, perhaps we could still add a comment to explain this? Something like this: `# Note that we need to create these functions in the helper function create_title_item_function to ensure that the <name> variable is bound properly. Without this intermediate helper, the wrong closure is taken, and the title string for *all* generated functions is set to the same value (namely the value it has at the end of the loop).`

doc/Tutorials.xml
```diff
@@ -264,4 +264,54 @@ function.
...
+Now, one can create a pdf and html document, like a package documentation out of it. Suppose the document above
+is saved as <F>worksheet.g</F>. Then, when GAP is started in the folder of this file, the command
+<Listing>
+AutoDocWorksheet( "worksheet.g" );
+</Listing>
+will create a subfolder called <F>doc</F> of the current directory in which it will create the documentation.
+There are several options to configure the output of the worksheet command, which are identical to the
+options of the <Ref Func="AutoDoc"/> command. It is even possible to test the examples in the worksheet using the maketest option
+from the AutoDoc command.
+Since the worksheets do not have a <F>PackageInfo.g</F> to extract information, all possible tags that &GAPDoc; supports for
```
fingolfin: perhaps insert a paragraph break above?

gap/AutoDocMainFunction.gd
```diff
@@ -44,9 +44,11 @@ DeclareGlobalFunction( "CreateMakeTest" );
...
-#!  This function works exactly like the AutoDoc command, except that no package is needed to create a
-#!  worksheet. It takes the sampe optional records as the AutoDoc-command, so please refer this command
-#!  for a full list.
-#!  It's only optional argument is a (list of) filenames, which are then scanned by the AutoDoc parser.
+#!  The intention of these function is to create stand-alone pdf and html files
+#!  using AutoDoc without having them associated to a package.
+#!  It uses the same optional records as the &AutoDoc; command itself, but instead of
+#!  a package name there should be a filename or a list of filenames containing AutoDoc
+#!  text from which the documents are created. Please see the &AutoDoc; command for more
+#!  information about this and have a look at the tutorial for a simple worksheet example.
```
fingolfin: You write `have a look at the tutorial` -- how about using a <Ref>  here to help the reader find it?

gap/Parser.gi
```diff
@@ -581,31 +581,28 @@ InstallGlobalFunction( AutoDoc_Parser_ReadFiles,
...
-        @URL := function( )
-            if not IsBound( tree!.TitlePage.TitleComment ) then
-                tree!.TitlePage.TitleComment := [ ];
-            fi;
-            tree!.content := tree!.TitlePage.TitleComment;
-
-            Add( tree, "<URL>" );
-            Add( tree, current_command[ 2 ] );
-            Add( tree, "</URL>" );
-        end
```
fingolfin: Seems you decided to remove @URL? But then it should also be removed from `doc/Comments.xml:432`, no?

sebasguts: Removed

fingolfin:
The changes in this PR look good to me now.

The commits themselves not so much. The commit message very often does not match the content. I think it would be better to either squash this into a single commit; and/or create new commits, e.g.

* one which removes @URL from both code and documentation;
* one which fixes the regression in `Parser.gi` by (re)introducing `create_title_item_function`; 
* one or two which fixes a typo and changes some code to use `AUTODOC_SetIfMissing`;
* one which adds the new worksheet documentation.

PR #99, opened 2016-02-23
Removed Index entry of bookname, closes #56

### LINKED ISSUE #56: Resolve FIXMEs in AutoDoc()
The FIXMEs introduced by b8b9aa09b should be resolved. 

- [x] The  first one introduces a weird entry to the index that seems to serve no purpose.

- [ ] The next one creates an DocumentationTree() object unconditionally. But shouldn't that only be created (and used!) if using AutoDoc mode, i.e. if "autodoc" is bound? Note that the tree object is later on also used later in code that is run even if autodoc is not bound (e.g. if is_worksheet is true or if scaffold.TitlePage is bound).

- [ ] The third issue is similar... It seems that maybe if "is_worksheet" is set to true, then autodoc and scaffold mode are implied. But this is not mentioned anywhere, nor checked nor enforced by the code. Thus if a user should forget to do that, they would get strange, not-helpful errors.

sebasguts:


fingolfin:
This PR only resolves half of #56

sebasguts:
You are right, got confused by the discussion. Made an update to the commit message

PR #105, opened 2016-02-28
Issues 103, 104, 106, 108

### LINKED ISSUE #104: Add .autodoc to list of automatically scanned extensions?
Perhaps we should autodoc to one of the two `AUTODOC_FindMatchingFiles` calls, namely the one which computes `autodoc.files`?

sebasguts:


fingolfin:
We should also mention this in the documentation, shouldn't we?

fingolfin:
I.e. in the docs for `AutoDoc()`, and perhaps also in the CHANGES file for the next release

gap/Magic.gd
```diff
@@ -303,6 +303,17 @@
 #!             <Br/>
 #!             <E>Default value: <C>[ ".", "gap", "lib", "examples", "examples/doc" ]</C>.</E>
 #!         </Item>
+#!         <Mark><A>gap_root_relative_path</A></Mark>
+#!         <Item>
+#!             Either true or a string representing a relative path. &GAPDoc;s MakeGAPDocDoc,
+#!             which is called to build the documentation, supports an optional parameter <A>gaproot</A>
+#!             that makes the links to other documentation entries relative. So this path, if present, should
```
fingolfin: "other documentation entries" should perhaps be "other manuals" or "external documentation (such as the GAP reference manual)"

gap/Magic.gd
```diff
@@ -303,6 +303,17 @@
...
+#!             be the relative position of the &GAP; root directory to the place where the documentation will be stored.
+#!             The default for this is not having a path, i.e., MakeGAPDocDoc will produce an HTML document with
```
fingolfin: Hmm, `The default for this is not having a path` sounds confusing. Perhaps: `By default, this value is not set and <C>MakeGAPDocDoc</C> produces references with absolute paths.`

CHANGES
```diff
@@ -4,6 +4,7 @@ XXX:
   - AutoDoc can now be instructed to invoke GAPDoc in such a way that links
     in the generated documentation to the GAP reference manual use relative
     paths. To do this, use TODO
```
fingolfin: The above incomplete CHANGES entry should be finished. Perhaps just remove the last incomplete sentence (the stuff before it should be enough, the user can find details in the documentation now).

gap/Magic.gd
```diff
@@ -303,6 +303,17 @@
 #!             <Br/>
 #!             <E>Default value: <C>[ ".", "gap", "lib", "examples", "examples/doc" ]</C>.</E>
 #!         </Item>
+#!         <Mark><A>gap_root_relative_path</A></Mark>
+#!         <Item>
+#!             Either true or a string representing a relative path. &GAPDoc;s MakeGAPDocDoc,
```
fingolfin: It could also be set to `false`, couldn't it? How about: `Either a boolean or a string...`

Also, `&GAPDoc;s MakeGAPDocDoc,` should use `<C>` -- and I am not sure whether I'd mention it here at all...

gap/Magic.gd
```diff
@@ -303,6 +303,17 @@
...
+#!             which is called to build the documentation, supports an optional parameter <A>gaproot</A>
+#!             that makes the links to other documentation entries relative. So this path, if present, should
+#!             be the relative position of the &GAP; root directory to the place where the documentation will be stored.
+#!             The default for this is not having a path, i.e., MakeGAPDocDoc will produce an HTML document with
+#!             absolute paths. If this value is set to true, the relative path <C>../../..</C> is used.<Br/>
+#!             There is also the possibility to set the option <C>relativePath</C> instead of using this value.
+#!             If both the option and the entry are set, the option is used.
+#!         </Item>
 #!
```
fingolfin: OK, so here is an alternative text for this:
```
#!             Either a boolean or a string containing a relative path.
#!             By default (if this option is not set, or is set to <K>false</K>),
#!             references in the generated documentation referring to external documentation
#!             (such as the GAP manual) are encoded using absolute paths.
#!             This is fine as long as the documentation stays on only a single
#!             computer, but is problematic when preparing documentation that should be
#!             used on multiple computers, e.g. when creating a distribution archive of
#!             a GAP package.<Br/>
#!             Thus, if a relative path is provided via this option (or if it is set to true,
#!             in which case the relative path <F>../../..</F> is used), then &AutoDoc;
#!             and &GAPDoc; attempt to replace all absolute paths in references to GAP
#!             manuals by paths based on this relative path.<P/>
#!             
#!             On a technical level, &AutoDoc; passes the relative path to the
#!             <A>gaproot</A> parameter of <Ref Func='MakeGAPDocDoc'
#!             BookName='gapdoc'/><P/>
#!             
#!             Additionally, you can also set the global option <C>relativePath</C>
#!             to a boolean or a string with the same effect (for details on global options,
#!             see <Ref Chap="Options Stack" BookName="ref"/>). If both this option
#!             record entry and the global option are set, then the global option takes
#!             precedence.
```

gap/Magic.gi
```diff
@@ -519,7 +519,11 @@ function( arg )
         # of the documentation are also in UTF-8 encoding, and may contain characters
         # not contained in the default Latin 1 encoding.
         SetGapDocLaTeXOptions( "utf8", gapdoc_latex_option_record );
-
+        
+        ## HACK: If there is an empty index, MakeGAPDocDoc throws an error when creating the pdf.
+        ## this addition prevents this by fake adding the index to the page number log.
+        Append( GAPDoc2LaTeXProcs.Tail, "\\immediate\\write\\pagenrlog{[\"Ind\", 0, 0], \\arabic{page},}\n" );
```
fingolfin: I'd mention #106 in the comment, e.g. `See issue #106` -- this way, one can quickly look up details.

gap/Magic.gd
fingolfin: Hmm, and we might want to *not* document `relativePath` at all, so that we can remove or at least rename it (it sticks out right now, because it uses CamelCase, where most other variables use snake_case)

gap/Magic.gd
```diff
@@ -303,6 +303,31 @@
...
+#!             On a technical level, &AutoDoc; passes the relative path to the
+#!             <A>gaproot</A> parameter of <Ref Func='MakeGAPDocDoc'
+#!             BookName='gapdoc'/><P/>
+#             
+#             Additionally, you can also set the global option <C>relativePath</C>
+#             to a boolean or a string with the same effect (for details on global options,
+#             see <Ref Chap="Options Stack" BookName="ref"/>). If both this option
+#             record entry and the global option are set, then the global option takes
+#             precedence.
+#!         </Item>
```
fingolfin: This is weird. To a casual observer, it seems as if the `!` is left out by accident. I'd really just remove this whole paragraph.

CHANGES
fingolfin: This still needs to be addressed.

PackageInfo.g
```diff
@@ -18,7 +18,7 @@ PackageName := "AutoDoc",
 Subtitle := "Generate documentation from GAP source code",
 
 Version := Maximum( [
-  "2016.02.23", ## Sebas' version
+  "2016.02.28", ## Sebas' version
```
fingolfin: Please also update this date, and then perhaps use `git rebase --ignore-date` to prevent your commit from looking like they are several days old, when in fact you just rewrote them? :-)

sebasguts:
Turns out, --ignore-date does not work together with -i

Issue #86, opened 2014-10-23
Entities in section titles lead to error


fingolfin:
The following leads to an error when used in an .autodoc file:

```
@Section What is &SingularInterface;?
```

This is because it produces this XML code:

```
<Section Label="Chapter_Introduction_Section_What_is_&SingularInterface;?">
```

Which is not valid...

fingolfin:
See also issue #127.

I assume this can be partially resolved with the help of `@SectionTitle` and friends. But the fact remains that we generate bad XML, and this can lead to very confusing GAPDoc errors. Better would be to catch any "invalid" input and immediately abort with an error. Of course *best* would be to just support this... :-).
In either case, we should also add test case to verify that the feature keeps working.

PR #111, opened 2016-03-02
Fix for #86, a doc improvement and better errors

fingolfin:


fingolfin:
Just added another minor usability improvement

Issue #101, opened 2016-02-24
Explain how to use AutoDoc in packages that already use GAPDoc


fingolfin:
Explain how to use AutoDoc in packages that already use GAPDoc, in more detail than Section 1.3.
- [x] a minimal example how to use `AutoDoc()` to compile your existing GAPDoc manual, nothing else. 
- [x] some examples showing the `gapdoc` options.
- [x] an explanation how to use `AutoDoc()` to generate a title page for use in a GAPDoc based manual (nothing else)
- [ ] examples for the fancier `scaffold` options
- [ ] finally, how to mix you GAPDoc manual with additional AutoDoc-style documentation comments.

This is motivated by #100

sebasguts:
The last item ''How to mix a GAPDoc manual with AutoDoc'' is already there, just hidden in 1.2. Should this be more prominent?

fingolfin:
I think this is pretty OK now, compared to when I wrote this issue originall. In any case, this is not concrete enough to be helpful. If somebody (including me) is unhappy with some aspects of the manual, they should either write a specific criticism for the current version of the manual, or submit a PR.

PR #113, opened 2016-03-03
Preserving leading spaces in code blocks, documentation.

sebasguts:
Some more work on the tutorials, specially for those mentioned in #101

### DIFF
--- a/doc/Comments.xml
+++ b/doc/Comments.xml
@@ -537,13 +537,12 @@ produces the following:
 </Section>
 
 <Section Label="MarkdownExtension">
-<Heading>Some syntax for &AutoDoc; text</Heading>
+<Heading>Markdown-like formatting of text in &AutoDoc;</Heading>
 
-&AutoDoc; offers some useful syntax extensions which can be used in &AutoDoc; plain text files
-and &AutoDoc; comments. The syntax is inspired by the Markdown language,
-but it does not implement all features, neither is it strict Markdown, but tries to
-be as straight as possible. The following constructions
-are possible right now:
+&AutoDoc; has some convenient ways to insert special format into text, like
+math formulas and lists. The syntax for them are inspired by Markdown and LaTeX,
+but do not follow them strictly. Neither are all features of the Markdown language
+supported. The following subsections describe what is possible.
 
 <Subsection Label="MarkdownExtensionList">
   <Heading>Lists</Heading>
--- a/doc/Tutorials.xml
+++ b/doc/Tutorials.xml
@@ -227,6 +227,60 @@ gap maketest.g
 test the examples from your manual.
 </Subsection>
 
+<Subsection Label="Tut:IntegrateExisting:GapDocOptions">
+<Heading>Setting different &GAPDoc; options</Heading>
+
+Sometimes, the default values for the &GAPDoc; command used by &AutoDoc;
+may not be suitable for your manual.
+<P/>
+ Suppose your main xml file is <E>not</E> named <F>PACKAGENAME.xml</F>, but rather something else, e.g. <F>main.xml</F>.
+Then you can tell &AutoDoc; to use this file as the main xml file via
+<Listing>
+LoadPackage( "AutoDoc" );
+AutoDoc( rec( gapdoc := rec( main := "main" ) ) );
+</Listing>
+<P/>
+As explained above, by default &AutoDoc; scans all <F>.g</F>, <F>.gd</F> and <F>.gi</F> files
+it can find inside of your package root directory, and in the subdirectories <F>gap</F>,
+<F>lib</F>, <F>examples</F> and <F>examples/doc</F> as well.
+If you keep source files with documentation in other directories,
+you can adjust the list of directories AutoDoc scans via the <C>scan_dirs</C> option.
+The following example illustrates this by instructing &AutoDoc; to only search in the subdirectory <F>package_sources</F>
+of the packages root directory.
+<Listing>
+LoadPackage( "AutoDoc" );
+AutoDoc( rec( gapdoc := rec( scan_dirs := [ "package_source" ] ) ) );
+</Listing>
+You can also specify an explicit list of files containing documentation,
+which will be searched in addition to any files located within the scan directories:
+<Listing>
+LoadPackage( "AutoDoc" );
+AutoDoc( rec( gapdoc := rec( files := [ "path/to/some/hidden/file.gds" ] ) ) );
+</Listing>
+Giving such a file does not prevent the standard <C>scan_dirs</C> from being scanned for
+other files.
+<P/>
+Next, &GAPDoc; supports the documentation to be built with relative paths.
+This means, links to manuals of other packages or the &GAP; library will
+not be absolute, but relative from your documentation. This can be particulary useful
+if you want to build a release tarball or move your &GAP; installation around later.
+Suppose you are starting &GAP; in the root path of your package as always,
+and the standard call of <Ref Func="AutoDoc"/> will then build the documentation in the <F>doc</F> subfolder of your package.
+From this folder, the gap root directory has the relative path <F>../../..</F>.
+Then you can enable the relative paths by
+<Listing>
+LoadPackage( "AutoDoc" );
+AutoDoc( rec( gapdoc := rec( gap_root_relative_path := "../../.." ) ) );
+</Listing>
+or, since <F>../../..</F> is the standard option for <C>gap_root_relative_path</C>, by
+<Listing>
+LoadPackage( "AutoDoc" );
+AutoDoc( rec( gapdoc := rec( gap_root_relative_path := true ) ) );
+</Listing>
+
+
+</Subsection>
+
 <Subsection Label="Tut:IntegrateExisting:TitlePage">
   <Heading>Using &AutoDoc;s scaffold for regenerating the title page</Heading>
 
--- a/doc/Tutorials.xml
+++ b/doc/Tutorials.xml
@@ -227,8 +227,26 @@ gap maketest.g
 test the examples from your manual.
 </Subsection>
 
-</Section>
+<Subsection Label="Tut:IntegrateExisting:TitlePage">
+  <Heading>Using &AutoDoc;s scaffold for regenerating the title page</Heading>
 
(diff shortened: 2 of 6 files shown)


doc/Comments.xml
```diff
@@ -537,13 +537,12 @@ produces the following:
...
 
-&AutoDoc; offers some useful syntax extensions which can be used in &AutoDoc; plain text files
-and &AutoDoc; comments. The syntax is inspired by the Markdown language,
-but it does not implement all features, neither is it strict Markdown, but tries to
-be as straight as possible. The following constructions
-are possible right now:
+&AutoDoc; has some convenient ways to insert special format into text, like
+math formulas and lists. The syntax for them are inspired by Markdown and LaTeX,
+but do not follow them strictly. Neither all features of the Markdown language
+are supported. The following subsections describe what is possible.
```
fingolfin: Should be "Neither are all", I think

doc/Tutorials.xml
```diff
@@ -181,12 +181,108 @@ There are various examples of packages using &AutoDoc; for only
...
+
+Suppose you already have a complete XML manual, with some main and title
+XML files and some documentation for operations distributed over
+all your <F>.g{,d,i}</F> files. Even then &AutoDoc; can do something for you.
+Suppose your main xml file is named <F>PACKAGENAME.xml</F> and is in the
+<F>/doc</F> subfolder of your package. Then calling
+<Listing>
+LoadPackage( "AutoDoc" );
+AutoDoc( : autodoc := false );
+</Listing>
```
fingolfin: Actually, I don't think you need any parameters, do you?

doc/Tutorials.xml
```diff
@@ -181,12 +181,108 @@ There are various examples of packages using &AutoDoc; for only
...
-TODO: Explain how to mix AutoDoc content with existing &GAPDoc; XML files.
--->
+<Subsection Label="Tut:IntegrateExisting:EverythingThere">
+  <Heading>Using &AutoDoc; on a complete &GAPDoc; manual</Heading>
+
+Suppose you already have a complete XML manual, with some main and title
+XML files and some documentation for operations distributed over
+all your <F>.g{,d,i}</F> files. Even then &AutoDoc; can do something for you.
+Suppose your main xml file is named <F>PACKAGENAME.xml</F> and is in the
+<F>/doc</F> subfolder of your package. Then calling
```
fingolfin: Should be "Then calling the following from a GAP started in the root directory of your package" or so

doc/Tutorials.xml
```diff
@@ -181,12 +181,108 @@ There are various examples of packages using &AutoDoc; for only
...
 
-<!--
-TODO: Explain how to mix AutoDoc content with existing &GAPDoc; XML files.
--->
+<Subsection Label="Tut:IntegrateExisting:EverythingThere">
+  <Heading>Using &AutoDoc; on a complete &GAPDoc; manual</Heading>
+
+Suppose you already have a complete XML manual, with some main and title
+XML files and some documentation for operations distributed over
+all your <F>.g{,d,i}</F> files. Even then &AutoDoc; can do something for you.
```
fingolfin: I am not sure people will understand this `.g{,d,i}` shorthand -- I'd spell it out as ".g, .gi and .gd files" or so

doc/Tutorials.xml
```diff
@@ -181,12 +181,108 @@ There are various examples of packages using &AutoDoc; for only
...
+Suppose your main xml file is named <F>PACKAGENAME.xml</F> and is in the
+<F>/doc</F> subfolder of your package. Then calling
+<Listing>
+LoadPackage( "AutoDoc" );
+AutoDoc( : autodoc := false );
+</Listing>
+will rebuild your manual. The main point in favor of doing is that you do
+not have to keep a list of all the <F>.g{,d,i}</F> files you have manual entries in.
```
fingolfin: Actually, I think there is a bit more to that... I was also working on some text for this, and this is what I wrote; perhaps you can salvage some of it? Note that I started from a slightly different angle (i.e. the assumption that there is an existing `makedocrel.g` which can be used as a starting point). Of course your approach is fine, too!
```
In order to build a &GAPDoc; based manual, one typically uses a
<F>makedoc.g</F> or <F>makedocrel.g</F> file which might
look something like this:

@BeginCode c1
LoadPackage( "GAPDoc" );
MakeGAPDocDoc( "doc", "manual", [ "../PackageInfo.g" ],
               "FooBar", "../../..", "MathJax");; 
CopyHTMLStyleFiles( "doc" );
GAPDocManualLab( "FooBar" );; 
@EndCode
@InsertCode c1

With &AutoDoc;, a roughly equivalent <F>makedoc.g</F> might look like this:

@BeginLog
if fail = LoadPackage("AutoDoc", ">= 2016.01.21") then
    Error("AutoDoc 2016.01.21 or newer is required");
fi;
AutoDoc(rec(gapdoc:=rec(main:="manual")));
@EndLog
```

doc/Tutorials.xml
```diff
@@ -181,12 +181,108 @@ There are various examples of packages using &AutoDoc; for only
...
+</Subsection>
 
-</Section>
+<Subsection Label="Tut:IntegrateExisting:GapDocOptions">
+<Heading>Setting different &GAPDoc; options</Heading>
 
+Sometimes, the default values for the &GAPDoc; command used by &AutoDoc;
+may not be suitable for your manual.
+<P/>
+Suppose you have your main xml file called <F>main.xml</F>. Then you can tell &AutoDoc;
```
fingolfin: Perhaps contrast this explicitly with the default:
`Suppose your main xml file is <E>not</E> named <F>PACKAGENAME.xml</F>, but rather something else, e.g. <F>main.xml</F>.`

doc/Tutorials.xml
```diff
@@ -181,12 +181,108 @@ There are various examples of packages using &AutoDoc; for only
...
+to use this file as the main xml file via
+<Listing>
+LoadPackage( "AutoDoc" );
+AutoDoc( : autodoc := false,
+           gapdoc := rec( main := "doc/main.xml" ) );
```
fingolfin: Hmm, isn't the `doc/` in there wrong? I think I tested it without that (see my example above). Best to double-check this.

And again, isn't the `autodoc:=false` superfluous?

fingolfin: BTW, I am more and more convinced that we should use the syntax `AutoDoc(rec(...))` in examples and `makedoc.g` etc. by default -- this way, it becomes possible to override AutoDoc settings before reading a `makedoc.g` file.... But we can also first discuss this elsewhere, and you leave yoru examples as they are... Just thinking out loud... :)

doc/Tutorials.xml
```diff
@@ -181,12 +181,108 @@ There are various examples of packages using &AutoDoc; for only
...
+</Listing>
+<P/>
+Also, it might happen that your <F>g.{,d,i}</F> files are not in somehow standard named
```
fingolfin: `are not in somehow standard named subdirectories` -- first off, this sounds wrong (grammar?), secondly, it's meaning seems unclear. Perhaps start by stating explicitly where GAP searches, and then covering the exception?

> As explained above, by default AutoDoc scans all .g, .gd and .gi files it can find inside of your package root directory, as well as the subdirectories gap, lib, examples and examples/doc. If you keep source files with documentation in other directories, you can adjust the list of directories AutoDoc scans via the scan_dirs option. The following example illustrates this by instructing AutoDoc to only search in the subdirectory package_sources of the package root directory.

doc/Tutorials.xml
```diff
@@ -181,12 +181,108 @@ There are various examples of packages using &AutoDoc; for only
...
+subdirectories of your package, but maybe in a folder called <F>package_sources</F>.
+Then you can tell &AutoDoc; to search this folder instead for the standard ones with
+<Listing>
+LoadPackage( "AutoDoc" );
+AutoDoc( : autodoc := false,
+           gapdoc := rec( scan_dirs := [ "package_source" ] ) );
+</Listing>
+It is also possible to give specific files, maybe hidden somewhere really deep in your
+packages folder structure.
```
fingolfin: > You can also specify an explici list of files containing documentation, which will be searched in addition to any files located within the scan directories:

doc/Tutorials.xml
```diff
@@ -181,12 +181,108 @@ There are various examples of packages using &AutoDoc; for only
...
+<Listing>
+LoadPackage( "AutoDoc" );
+AutoDoc( : autodoc := false,
+           gapdoc := rec( files := [ "path/to/some/hidden/file.gds" ] ) );
+</Listing>
+Giving such a file does not prevent the standard <C>scan_dirs</C> from being scanned for
+other files.
+<P/>
+Next, &GAPDoc; supports the documentation to be build with relative paths.
```
fingolfin: build -> built

doc/Tutorials.xml
```diff
@@ -181,12 +181,108 @@ There are various examples of packages using &AutoDoc; for only
...
+This means, links to documentations of other packages or the &GAP; library will
```
fingolfin: documentations -> documentation (or say "manuals")

doc/Tutorials.xml
```diff
@@ -181,12 +181,108 @@ There are various examples of packages using &AutoDoc; for only
...
+not be absolute, but relative from your documentation. This can be particulary useful
+if you want to build a release tarball or move your &GAP; installation around later.
+Suppose you are building your documentation in the <F>doc</F> subfolder of your package
```
fingolfin: I find this sentence confusing. It sounds as if you are talking about invoking `AutoDoc()` from within the `doc/` subdirectory, something that normally is never done. So I am not quite sure what you mean here...

fingolfin:
Overall: Good job! Just a bunch of nitpicks in the documentation ;-)

doc/Tutorials.xml
sebasguts: I think this is a good idea, so making it possible to overwrite these parameters in scripts. I change those in all examples

sebasguts:
Thanks. Hopefully most of the stuff you mentioned is gone now.

fingolfin:
Thanks. I merged this, and added a few edits atop.

PR #121, opened 2016-11-28
Rewrite entities support and document it

### LINKED ISSUE #120: Allow setting arbitrary entities, and document how it is done
Right now, there is an undocumented way to set custom entities via `AutoDoc`
```gap
AutoDoc(rec( 
    autodoc := true,
    scaffold := rec(
        includes := [
            "Tutorials.xml",
            "Comments.xml"
            ],
        entities := [
            "SomePackage",  # declares entity "SomePackage" with value "<Package>SomePackage</Package>"
            [ "ent", "val" ],  # declares entity "ent" with value "val"
            "<!ENTITY RELEASEYEAR \"2015\">",  # entity declaration is copied varbatim
        ],
    )
));
```
We did not document this yet, though, partially (I think) because it was not completely clear whether that's the way to go. 

And indeed, I don't really like it. Instead, I propose that we do what `GAPDoc` does (e.g. for `ParseTreeXMLString`): Namely, pass entity declaration as a record, where the keys are the entity names, and the values are used as entity content. The values could be limited to strings, though I think it would be OK to also accept other things and automatically invoke `String` on them...

Also, let's get rid of the implicit `<Package>` -- it is not much of a burden to enter this once, but adding it automatically makes it yet again harder to learn how things work.

So, the above example would become this:
```gap
AutoDoc(rec( 
    autodoc := true,
    scaffold := rec(
        includes := [
            "Tutorials.xml",
            "Comments.xml"
            ],
        entities := rec(
            SomePackage := "<Package>SomePackage</Package>",
            ent := "val",
            RELEASEYEAR := "2015",
        ],
    )
));
```

BTW, our entity handling should of course correctly deal with quotes `"` or `'` inside the entity value.

Finally, we would of course keep suppoting the old-style entities declaration for backwards compatibility. Perhaps issue a warning about it, though?

See also issue #100 

UPDATE: My description was incorrect when it comes to package names, which is now corrected.

fingolfin:
Resolves #120 

The old entities code is kept for backwards compatibilit, but is intentionally left undocumented

A potential enhancement to this would be to print a warning if the entities are specifiedas a list.

### DIFF
--- a/gap/AutoDocMainFunction.gi
+++ b/gap/AutoDocMainFunction.gi
@@ -77,19 +77,24 @@ end );
 ##
 InstallGlobalFunction( CreateMainPage,
   function( book_name, dir, opt )
-    local filename, filestream, i;
+    local filename, filestream, i, ent, val;
 
     if IsString(dir) then
         dir := Directory(dir);
     fi;
 
     if not IsBound( opt.entities ) then
-        opt.entities := [];
+        opt.entities := rec();
     fi;
 
     # add book_name unconditionally to the list of entities
-    # FIXME: stop doing that, to allow package authors to define this entity differently?
-    Add( opt.entities, book_name );
+    if IsRecord( opt.entities ) then
+        if not IsBound(opt.entities.(book_name)) then
+            opt.entities.(book_name) := Concatenation( "<Package>", book_name, "</Package>" );
+        fi;
+    else
+        Add( opt.entities, book_name );
+    fi;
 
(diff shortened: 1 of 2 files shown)


fingolfin:
@sebasguts If I don't hear any objections from you soon, I'll just merge (or rather, "rebase", yay to GitHub improvements) this, and make a new AutoDoc release soon after.

sebasguts:
I do not see any objections to this, so we can merge it. Are you going to create the new release?

PR #125, opened 2017-01-27
Autodoc did not scan KeyDependendOperations filters correct.

### LINKED ISSUE #124: Problem in documenting key dependent operations.
Hallo :) 

When I try to document lets say:

#! @Description
#! This computes .....
#! @Arguments C, n
#! @Returns a morphism
KeyDependentOperation( "CertainCycle", IsChainOrCochainComplex, IsInt, ReturnTrue );

In the manual.pdf one finds

**CertainCycle (for IsChainOrCochainComplex, IsChainOrCochainComplex)**
CertainCycle(C, n)                                                                                                         (operation)
Returns: a morphism
This computes ...

The second argument is supposed to be "IsInt".

Best wishes

sebasguts:
Closes #124.

### DIFF
--- a/gap/Parser.gi
+++ b/gap/Parser.gi
@@ -249,6 +249,9 @@ InstallGlobalFunction( AutoDoc_Parser_ReadFiles,
                 if current_item!.tester_names = fail and StripBeginEnd( filter_string, " " ) <> "for" then
                     current_item!.tester_names := filter_string;
                 fi;
+                if StripBeginEnd( filter_string, " " ) = "for" then
+                    has_filters := "empty_argument_list";
+                fi;
                 ##Adjust arguments
                 if not IsBound( current_item!.arguments ) then
                     if IsInt( has_filters ) then
@@ -264,6 +267,8 @@ InstallGlobalFunction( AutoDoc_Parser_ReadFiles,
                         else
                             current_item!.arguments := JoinStringsWithSeparator( current_item!.arguments, "," );
                         fi;
+                    elif has_filters = "empty_argument_list" then
+                        current_item!.arguments := "";
                     fi;
                 fi;
             fi;
--- a/gap/Parser.gi
+++ b/gap/Parser.gi
@@ -61,6 +61,13 @@ InstallGlobalFunction( AutoDoc_Type_Of_Item,
     elif PositionSublist( type, "DeclareOperation" ) <> fail then
         entries := [ "Oper", "methods" ];
         has_filters := "List";
+    elif PositionSublist( type, "DeclareConstructor" ) <> fail then
+        ## FIXME: there should be a Constructor tag, but it is unfortunately not possible, since GAPDoc
+        ##        does not offer such a tag. Issue for this is filed here:
+        ##        https://github.com/frankluebeck/GAPDoc/issues/23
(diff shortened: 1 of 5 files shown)


gap/Parser.gi
```diff
@@ -211,7 +212,16 @@ InstallGlobalFunction( AutoDoc_Parser_ReadFiles,
                         current_line := ReadLineWithLineCount( filestream );
                         NormalizeWhitespace( current_line );
                     od;
-                    Append( filter_string, StripBeginEnd( current_line{ [ 1 .. Minimum( [ PositionSublist( current_line, "," ), PositionSublist( current_line, ");" ) ] ) - 1 ] }, " " ) );
+                    current_line_positition_for_filter := Minimum( [ PositionSublist( current_line, "," ), PositionSublist( current_line, ");" ) ] ) - 1;
```
fingolfin: Could you please add some comments here, so that on has at least an idea what this is all about? If I didn't have the tag line, I would never have guessed its purpose.

gap/Parser.gi
```diff
@@ -211,7 +212,16 @@ InstallGlobalFunction( AutoDoc_Parser_ReadFiles,
                         current_line := ReadLineWithLineCount( filestream );
                         NormalizeWhitespace( current_line );
                     od;
```
fingolfin: It seems that further up, you explicitly check for `KeyDependentOperation`. But then mix it with a check for `Declare*`. How about remembering the result of that check, and then using it here? That would result in more robust code, wouldn't it? As you could use that to decide whether to look for `);` or `,`, no? It would also automatically make clear what's going on.

gap/Parser.gi
```diff
@@ -211,7 +212,16 @@ InstallGlobalFunction( AutoDoc_Parser_ReadFiles,
...
-                    Append( filter_string, StripBeginEnd( current_line{ [ 1 .. Minimum( [ PositionSublist( current_line, "," ), PositionSublist( current_line, ");" ) ] ) - 1 ] }, " " ) );
+                    current_line_positition_for_filter := Minimum( [ PositionSublist( current_line, "," ), PositionSublist( current_line, ");" ) ] ) - 1;
+                    Append( filter_string, StripBeginEnd( current_line{ [ 1 .. current_line_positition_for_filter ] }, " " ) );
+                    current_line := current_line{[ current_line_positition_for_filter + 1 .. Length( current_line ) ]};
+                    if current_line[ 1 ] = ',' then
+                        current_line := current_line{[ 2 .. Length( current_line ) ]};
+                    elif current_line[ 1 ] = ')' then
+                        current_line := current_line{[ 3 .. Length( current_line ) ]};
+                    fi;
+                    ## FIXME: This elif case should never be reached. Instead there should be a break; if ); is reached.
```
fingolfin: Hmm, so are you saying here "it never is reached in all my tests, I only left it here out of paranoia" ?
Or are you saying: "The code should be changed so that it never is reached" ?

sebasguts: It means that by logic this can never be reached. I still left it there out of paranoia.

sebasguts: Completely wrong, I am sorry. This is just there because there is no break statement. I guess I can remove this fixme.

PR #129, opened 2017-08-03
Several improvements

sebasguts:
* Correct scanning of ´KeyDependendOperation`
* No label for functions with no filters
* No argument for functions with no argument
* Documentation of `DeclareConstructor`
* No return statement printed if no `@Returns` is given
* ChapterLabel, SectionLabel, SubsectionLabel
* Listing for GAPDoc example style
* Fixup of example doc

### DIFF
--- a/doc/Comments.xml
+++ b/doc/Comments.xml
@@ -138,6 +138,7 @@ text in your documentation, examples, mathematical chapters, etc..
 
 <Subsection Label="@Chapter">
 <Index Key="@Chapter"><C>@Chapter</C></Index>
+<Index Key="@ChapterLabel"><C>@ChapterLabel</C></Index>
 <Heading>@Chapter <A>name</A></Heading>
 Sets a chapter, all functions without separate info will be added to this chapter.
 Also all text comments, i.e. lines that begin with #! without a command, and which do not
@@ -150,10 +151,18 @@ Example:
 #!  This is my chapter.
 #!  I document my stuff in it.
 ]]></Listing>
+
+The <C>@ChapterLabel</C> <A>label</A> command
+can be used to set the label of the chapter to <C>Chapter_</C><A>label</A>.
+
+Additionally, the chapter will be stored as <C>_Chapter_</C><A>label</A><C>.xml</C>.
+
+
 </Subsection>
 
 <Subsection Label="@Section">
 <Index Key="@Section"><C>@Section</C></Index>
+<Index Key="@SectionLabel"><C>@SectionLabel</C></Index>
 <Heading>@Section <A>name</A></Heading>
 Sets a section like chapter sets a chapter.
 
@@ -161,6 +170,9 @@ Sets a section like chapter sets a chapter.
 #! @Section My first manual section
 #!  In this section I am going to document my first method.
 ]]></Listing>
+
+The <C>@SectionLabel</C> <A>label</A> command
+can be used to set the label of the section to <C>Section_</C><A>label</A>.
 </Subsection>
 
 <Subsection Label="@EndSection">
@@ -178,6 +190,7 @@ opening it might cause unexpected errors.
 
 <Subsection Label="@Subsection">
 <Index Key="@Subsection"><C>@Subsection</C></Index>
+<Index Key="@SubsectionLabel"><C>@SubsectionLabel</C></Index>
 <Heading>@Subsection <A>name</A></Heading>
 Sets a subsection like chapter sets a chapter.
 
@@ -185,6 +198,10 @@ Sets a subsection like chapter sets a chapter.
 #! @Subsection My first manual subsection
 #!  In this subsection I am going to document my first example.
 ]]></Listing>
+
+The <C>@SubsectionLabel</C> <A>label</A> command
+can be used to set the label of the subsection to <C>Subsection_</C><A>label</A>.
+
 </Subsection>
 
 <Subsection Label="@EndSubsection">
--- a/doc/Comments.xml
+++ b/doc/Comments.xml
@@ -314,11 +314,49 @@ Order(S5);
 ]]></Listing>
 </Subsection>
 
+
+<Subsection Label="@ExampleSession">
+<Index Key="@ExampleSession"><C>@ExampleSession, @BeginExampleSession, and @EndExampleSession</C></Index>
+<Heading>@ExampleSession and @EndExampleSession</Heading>
+@ExampleSession inserts an example into the manual, but in a different syntax.
+For example, consider
+<Listing><![CDATA[
+#! @BeginExample
+S5 := SymmetricGroup(5);
+#! Sym( [ 1 .. 5 ] )
+Order(S5);
+#! 120
+#! @EndExample
+]]></Listing>
+As you can see, the commands are not commented and hence are executed when the file containing
+the example block is read. To insert examples directly into source code files, one can instead
+use @ExampleSession:
+<Listing><![CDATA[
+#! @ExampleSession
+#! gap> S5 := SymmetricGroup(5);
+#! Sym( [ 1 .. 5 ] )
+#! gap> Order(S5);
+#! 120
+#! @EndExampleSession
+]]></Listing>
+It inserts an example into the manual just as @Example would do, but all lines are commented
+and therefore not executed when the file is read.
+</Subsection>
+
+
+
 <Subsection Label="@BeginLog">
 <Index Key="@BeginLog"><C>@BeginLog and @EndLog</C></Index>
 <Heading>@BeginLog and @EndLog</Heading>
 Works just like the @BeginExample command, but the example
 will not be tested. See the GAPDoc manual for more information.
+</Subsection>
+
+<Subsection Label="@BeginLogSession">
+<Index Key="@BeginLogSession"><C>@BeginLogSession and @EndLogSession</C></Index>
+<Heading>@BeginLogSession and @EndLogSession</Heading>
+Works just like the @BeginExampleSession command, but the example
+will not be tested if manual examples are run. See the GAPDoc manual for more information.
 </Subsection>
 
 <Subsection Label="@DoNotReadRestOfFile">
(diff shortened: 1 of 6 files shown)


doc/Comments.xml
```diff
@@ -150,17 +151,31 @@ Example:
 #!  This is my chapter.
 #!  I document my stuff in it.
 ]]></Listing>
+
+Labels of chapters (for referencing purposes) are <C>Chapter_</C><A>name</A>, where
+all spaces in <A>name</A> are replaced by underscores. The <C>@ChapterLabel</C> <A>label</A> command
+can be used to set a custom label. This will set the label of the chapter to <C>Chapter_</C><A>label</A>.
```
fingolfin: "This will" is unclear -- what is "this"? Perhaps write this:

"You can set a custom label by using the <C>@ChapterLabel</C><A>label</A> command, which will set the label of the chapter to <C>Chapter_</C><A>label</A>."

doc/Comments.xml
```diff
@@ -150,17 +151,31 @@ Example:
...
+
+Additionally, the chapter will be stored as <C>_Chapter_</C><A>label</A><C>.xml</C>.
```
fingolfin: This is a remark about an internal feature -- and I don't think we explain anywhere right now how AutoDoc works (i.e. that it generates GAPDoc XML files and then feeds them into GAPDoc), do we? So the above sentence is only useful for the handful of experts who know the AutoDoc internals, and confusing to the rest.

So either remove it; or, better (on the long run): add a section somewhere which sketches how AutoDoc works (or perhaps it is there and I forgot about it; didn't even look now). In that case, I'd write something like this: "As described in REF, AutoDoc prouces GAPDoc XML files as an intermediate step. Each chapter is placed into a separate XML file, with the name <C>_Chapter_</C><A>label</A><C>.xml</C>."

doc/Comments.xml
```diff
@@ -150,17 +151,31 @@ Example:
...
 
 <Listing><![CDATA[
 #! @Section My first manual section
 #!  In this section I am going to document my first method.
 ]]></Listing>
+
+Labels of sections (for referencing purposes) are <C>Chapter_chaptername_Section_</C><A>name</A>, where
+all spaces in <A>name</A> are replaced by underscores. The <C>@SectionLabel</C> <A>label</A> command
+can be used to set a custom label. This will set the label of the section to <C>Section_</C><A>label</A>.
+
```
fingolfin: Same comments as for chapters apply

doc/Comments.xml
```diff
@@ -283,8 +304,9 @@ Resets the current level to 0.
 <Subsection Label="@BeginExample">
 <Index Key="@BeginExample"><C>@BeginExample and @EndExample</C></Index>
 <Heading>@BeginExample and @EndExample</Heading>
-@BeginExample inserts an example into the manual. The syntax is like the example environment
-in GAPDoc. This examples can be tested by GAPDoc, and also stay readable by GAP.
+@BeginExample inserts an example into the manual, with a different syntax to create a GAP readable
```
fingolfin: "with a different syntax" -- different from what?

doc/Comments.xml
```diff
@@ -297,6 +319,25 @@ Order(S5);
...
 </Subsection>
 
+
+<Subsection Label="@Listing">
+<Index Key="@Listing"><C>@Listing, @BeginListing, and @EndListing</C></Index>
+<Heading>@BeginListing <A>tested</A> and @EndListing</Heading>
+@BeginListing inserts an example into the manual. Opposed to @Example they have
+the same Syntax as GAPDoc examples and each line has to begin with a GAPDoc comment.
+@EndListing ends the example block.
+If <A>tested</A> is <C>false</C>, the example will not be tested if the manual examples are tested.
```
fingolfin: `tested` to me "feels" wrong... I"d prefer "test" or "addToTests" or "includeInTests" or so... Hrm...

Also, I would make this optional -- by default, all listings / examples are tested, but I can mark some of them so that they will be not tested.

Indeed it might be easier if you did not allow a boolean here, but rather `@Listing` and `@Listing notest` or so.

doc/Comments.xml
```diff
@@ -297,6 +319,25 @@ Order(S5);
 ]]></Listing>
 </Subsection>
 
+
+<Subsection Label="@Listing">
```
fingolfin: I looked at the GAPDoc documentation, and it offers listings, examples, and logs. We should *not* confuses people more by calling GAPDoc examples "listings". Let's have @Listing produce a GAPDoc List, and lts' have `@Log` begin a GAPDoc log. And if we can't have @Example begin a GAPDoc example (sad :-/), then we pick a different name for that. Don't have a bright idea for that right now, though... hrm

fingolfin: Perhaps just use @Log with a `include_in_tests` flag... but that also is confusing to people who use GAPDoc.
Of course people who do not use GAPDoc would be less affected...

Still, the very clear choice for this would be @Example, it's unfortunate this has been taken already. So... perhaps `@Demo`?  (I'd have used that for what is now called @Example, but I understand it's a bit late now to make changes here)

doc/Comments.xml
sebasguts: The testing argument is already optional, but I will change it to not be a boolean.

sebasguts:
Regarding the `@Example` issue, I can right now think of the following solution:
* Calling the new example way `@GAPDocExample`, which is kind of strange.
* Calling the old example `@AutoDocExample`, which will break tons of existing examples, so this has to be done very carefully. In principle, it should be possible to implement a heuristic to see if `@Example` is used for old or new style examples. We could then for a time use the heuristic, together with the printing of a warning message.
* Introducing a v2 of the language, which can be specified in either
  + the package info;
  + the `AutoDoc` command
  + using an `@Version` tag.

* A mix of one and two.

I do not particulary like any of those. The first is probably the easiest for now.

fingolfin:
Elswhere, you suggested `@SessionExample`... How about `@ExampleSession` then (and perhaps the "AutoDoc format v2" one day in the future?)

doc/Comments.xml
```diff
@@ -314,13 +314,40 @@ Order(S5);
 ]]></Listing>
 </Subsection>
 
+
+<Subsection Label="@ExampleSession">
```
fingolfin: The `ExampleSession` commit looks fine.

doc/Comments.xml
```diff
@@ -300,9 +300,11 @@ Resets the current level to 0.
 <Subsection Label="@BeginExample">
 <Index Key="@BeginExample"><C>@BeginExample and @EndExample</C></Index>
 <Heading>@BeginExample and @EndExample</Heading>
-@BeginExample inserts an example into the manual. The syntax is like the example environment
```
fingolfin: Typo in commit message: "documtentation"

doc/Comments.xml
```diff
@@ -300,9 +300,11 @@ Resets the current level to 0.
...
-in GAPDoc. This examples can be tested by GAPDoc, and also stay readable by GAP.
-The GAP prompt is added by AutoDoc.
+@BeginExample inserts an example into the manual. The syntax for examples is different from
+GAPDocs example syntax in order to have a
```
fingolfin: "in order to have a file that contains the example and is GAP readable" is weird. Perhaps "in order to allow creating files which contain an example but still are GAP readable" ?

doc/Comments.xml
```diff
@@ -300,9 +300,11 @@ Resets the current level to 0.
...
+file that contains the example and is GAP readable. GAP commands are not preceded by a comment, while output has to be preceded by an
+&AutoDoc; comment. So examples can be tested by GAPDoc, and also stay readable by GAP.
```
fingolfin: Of course then the second sentences becomes a bit redundant. So reword that, too?

doc/Comments.xml
```diff
@@ -300,9 +300,11 @@ Resets the current level to 0.
...
+The GAP prompt for the display in the manual is added by AutoDoc.
```
fingolfin: I find this a bit hard to understand. Perhaps remove this, instead add: "For example, consider this: [ Example block follows]. As you can see, the commands are not commented out, and hence are executed if the file containing this example block is read. At the same time, when building the documentation, this inserts an example equivalent to what this would produce: [ insert equivalent Begin/EndExampleSesson ]"

doc/Comments.xml
fingolfin: Also, this doesn't just "fix" the documentation... Perhaps: "Improve documentation for @Example"

gap/Markdown.gi
```diff
@@ -25,13 +25,22 @@ InstallGlobalFunction( CONVERT_LIST_OF_STRINGS_IN_MARKDOWN_TO_GAPDOC_XML,
...
           current_position, already_in_list, command_list_with_translation, beginning,
-          commands, position_of_command, insert, beginning_whitespaces, temp, string_list_temp, skipped;
+          commands, position_of_command, insert, beginning_whitespaces, temp, string_list_temp, skipped,
+          already_inserted_paragraph;
 
-    ## Check for paragraphs by making an empty string into <br/>
-    for i in [ 2 .. Length( string_list ) - 1 ] do
-        if string_list[ i ] = "" then
-            string_list[ i ] := "<Br/>";
+    ## Check for paragraphs by making an empty string into <P/>
```
fingolfin: making -> turning

Otherwise the commit "Made a #! followed by nothing parsed as <P/>."  is fine, assuming it works as intended -- perhaps we could add an implicit test for it, by immediately making use of this feature in the AutoDoc manual?

gap/Parser.gi
```diff
@@ -249,6 +249,9 @@ InstallGlobalFunction( AutoDoc_Parser_ReadFiles,
                 if current_item!.tester_names = fail and StripBeginEnd( filter_string, " " ) <> "for" then
                     current_item!.tester_names := filter_string;
                 fi;
+                if StripBeginEnd( filter_string, " " ) = "for" then
+                    has_filters := "empty_argument_list";
+                fi;
```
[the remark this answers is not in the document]
sebasguts: Before, if no `@Arguments` was given, `arg` was inserted automatically. This is fixed now. We could think about inserting `arg...`.

PR #135, opened 2017-09-07
Declare constructor

sebasguts:
Scan `DeclareConstructor` and automatically put constructor ManItem.

### DIFF
--- a/gap/Parser.gi
+++ b/gap/Parser.gi
@@ -38,6 +38,10 @@ InstallGlobalFunction( Scan_for_AutoDoc_Part,
     return [ command, argument ];
 end );
 
+
+BindGlobal( "AutoDoc_PrintWarningForConstructor",
+            AutoDoc_CreatePrintOnceFunction( "Installed GAPDoc version does not support constructors" ) );
+
 ##
 InstallGlobalFunction( AutoDoc_Type_Of_Item,
   function( current_item, type, default_chapter_data )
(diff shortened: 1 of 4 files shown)


PackageInfo.g
```diff
@@ -93,7 +93,7 @@ PackageDoc := rec(
 
 Dependencies := rec(
   GAP := ">= 4.5",
-  NeededOtherPackages := [ [ "GAPDoc", ">= 1.5" ] ],
+  NeededOtherPackages := [ [ "GAPDoc", ">= 1.6.1" ] ],
```
fingolfin: GAPDoc 1.6.1 has not yet been released, so I don't think we should merge this before.

gap/Parser.gi
```diff
@@ -62,11 +62,7 @@ InstallGlobalFunction( AutoDoc_Type_Of_Item,
         entries := [ "Oper", "methods" ];
         has_filters := "List";
     elif PositionSublist( type, "DeclareConstructor" ) <> fail then
-        ## FIXME: there should be a Constructor tag, but it is unfortunately not possible, since GAPDoc
-        ##        does not offer such a tag. Issue for this is filed here:
-        ##        https://github.com/frankluebeck/GAPDoc/issues/23
-        ##        Once this is fixed, the next line needs to be changed accordingly.
-        entries := [ "Oper", "methods" ];
+        entries := [ "Constr", "methods" ];
```
fingolfin: If you are eager to merge this now, perhaps do something as in this pseudo code:
```
  if GAPDoc >= 1.6.1 then
      entries := [ "Constr", "methods" ];
  else
     Print("Warning: Installed GAPDoc version does not support constructors");
     entries := [ "Oper", "methods" ];
  fi;
```
or even change the `Print` to an `Error`. In either case, you can drop the requirement for GAPDoc 1.6.1 from `PackageInfo.g`

sebasguts: Currently, it is fine as it is. It prints constructors as Operations, as it was before. So there is no need to merge this until the new GAPDoc is released.

PR #140, opened 2017-10-03
Spacing in new @ExampleSession environment:

sebasguts:
Since GAPDoc only highlights the examples properly if there is no space in front
of the gap> prompt, the standard is now that one character after the AutoDoc comments is
removed.

### DIFF
--- a/doc/Comments.xml
+++ b/doc/Comments.xml
@@ -343,6 +343,12 @@ use @ExampleSession:
 ]]></Listing>
 It inserts an example into the manual just as @Example would do, but all lines are commented
 and therefore not executed when the file is read.
+All lines that should be part of the example displayed in the manual
+have to start with an &AutoDoc; comment (<C>#!</C>).
+The comment will be removed, and, if the following character is a space,
+this space will also be removed. There is never more than one space removed.
+To ensure examples are correctly colored in the manual, 
+there should be exactly one space between <C>#!</C> and the <C>gap</C> prompt.
 </Subsection>
 
 
--- a/gap/Parser.gi
+++ b/gap/Parser.gi
@@ -437,6 +437,9 @@ InstallGlobalFunction( AutoDoc_Parser_ReadFiles,
             temp_pos_comment := PositionSublist( temp_curr_line, "#!" );
             if temp_pos_comment <> fail then
                 temp_curr_line := temp_curr_line{[ temp_pos_comment + 2 .. Length( temp_curr_line ) ]};
+                if Length( temp_curr_line ) >= 1 and temp_curr_line[ 1 ] = ' ' then
+                    Remove( temp_curr_line, 1 );
+                fi;
                 Add( temp_string_list, temp_curr_line );
             fi;
         od;


doc/Comments.xml
```diff
@@ -343,6 +343,10 @@ use @ExampleSession:
 ]]></Listing>
 It inserts an example into the manual just as @Example would do, but all lines are commented
 and therefore not executed when the file is read.
+All lines have to start with an &AutoDoc; comment
+(<C>#!</C>) followed by exactly one space. Both the comment and the space are removed for the manual.
```
fingolfin: Really *exactly one* space, not "at least one"? What about this:
```
#! gap> Print("  test\n");
#!   test
```

doc/Comments.xml
```diff
@@ -343,6 +343,10 @@ use @ExampleSession:
...
+If there are more spaces between the <C>gap</C> promt and the comment the example will not
+be colored properly in the manual.
```
fingolfin: typo: `promt` (also in the commit message)

doc/Comments.xml
sebasguts: Bad wording, sorry. Of course, one can put more spaces. But *exactly one* is removed, to allow what you typed.

doc/Comments.xml
```diff
@@ -343,6 +343,10 @@ use @ExampleSession:
 ]]></Listing>
 It inserts an example into the manual just as @Example would do, but all lines are commented
 and therefore not executed when the file is read.
+All lines have to start with an &AutoDoc; comment
+(<C>#!</C>). The comment and the following character are removed for the manual.
```
fingolfin: So... we now accept everything? Even `#!TextWithNoSpace`, and then truncate that to `extWithNoSpace`?

doc/Comments.xml
```diff
@@ -343,6 +343,10 @@ use @ExampleSession:
 ]]></Listing>
 It inserts an example into the manual just as @Example would do, but all lines are commented
 and therefore not executed when the file is read.
+All lines have to start with an &AutoDoc; comment
```
fingolfin: Super nitpick, but why the weird line wrapping, and why end the line here? You move the last "word" of the sentence to a new line, when you then have a muuuch longer line coming next...

doc/Comments.xml
```diff
@@ -343,6 +343,10 @@ use @ExampleSession:
...
+(<C>#!</C>). The comment and the following character are removed for the manual.
+To have examples correctly colored in the manual, there should be exactly one space
+between <C>#!</C> and the <C>gap</C> prompt.
```
fingolfin: Perhaps: "To ensure examples are correctly colored ..."

doc/Comments.xml
sebasguts: Yes. How should we decide what is the right behaviour? Only remove spaces? We could also do this.

doc/Comments.xml
```diff
@@ -343,6 +343,11 @@ use @ExampleSession:
 ]]></Listing>
 It inserts an example into the manual just as @Example would do, but all lines are commented
 and therefore not executed when the file is read.
+All lines have to start with an &AutoDoc; comment (<C>#!</C>).
```
fingolfin: Hrm, what does "All lines" here mean? Clearly in the above example, not all lines start with `#!`

doc/Comments.xml
```diff
@@ -343,6 +343,11 @@ use @ExampleSession:
...
+The comment will be removed. Also, if the following character is a space,
+it will also be removed. There is never more than one character removed.
```
fingolfin: two times "also" in one sentence: not nice.

Also: you remove 2 or 3 characters, so what do you mean with "never more than one character removed" ? Perhaps "space" instead of "character?

PR #142, opened 2017-10-07
Make copy of string before printing string,

### LINKED ISSUE #141: Markdown-lists don't work properly in chunks
Consider the following code:
```
#! @BeginChunk List
#! * Lorem ipsum dolor sit amet, consetetur sadipscing elitr, 
#!   sed diam nonumy eirmod tempor invidunt ut labore et 
#!   dolore magna aliquyam erat, sed diam voluptua.
#! * At vero eos et accusam et justo duo dolores et ea rebum. 
#!   Stet clita kasd gubergren, no sea takimata sanctus est 
#!   Lorem ipsum dolor sit amet.
#! @EndChunk
```
```
#! @InsertChunk List
```
This will result in a broken list (only the lines that have a * will be part of a list). Usually AutoDoc recognizes that the next lines are still part of the list and formats correctly.

If you don't manually wrap the lines, it works. Using the GAPDoc-syntax
```
<List>
  <Item>bla</Item>
</List>
```
also works fine.

sebasguts:
and do not use in-place NormalizeWhitespace in printing of string
since normalizing the string can cause markdown parser to fail the
second time a chunk is inserted. Closes #141.

### DIFF
--- a/gap/DocumentationTree.gi
+++ b/gap/DocumentationTree.gi
@@ -518,7 +518,7 @@ InstallMethod( WriteDocumentation, [ IsList, IsStream ],
     current_string_list := [ ];
     for i in [ 1 .. Length( node_list ) ] do
         if IsString( node_list[ i ] ) then
-            Add( current_string_list, node_list[ i ] );
+            Add( current_string_list, ShallowCopy( node_list[ i ] ) );
         else
             if current_string_list <> [ ] then
                 current_string_list := CONVERT_LIST_OF_STRINGS_IN_MARKDOWN_TO_GAPDOC_XML( current_string_list );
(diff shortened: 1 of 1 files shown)


gap/DocumentationTree.gi
```diff
@@ -541,7 +541,7 @@ InstallMethod( WriteDocumentation, [ IsString, IsStream ],
     ## In case the list is empty, do nothing.
     ## Once the empty string = empty list bug is fixed,
```
fingolfin: Wait, what bug is that? Does it have an open issue in this repo? If not, perhaps add one? And then reference the number of that issue in that comment?

sebasguts: In GAP, it is
```
gap> [ ] = "";
true
```
I do not consider this a bug anymore, so this comment should be updated to something like
> Since it is `[ ] = ""` ...
However, should this be done for this PR? The comment is old.

fingolfin: It can be done in this PR or in a new one; I simply stumbled over this, trying to understand what is going on in this PR (which I still don't really know, but that's OK :-).

sebasguts:
I resolved the conflict. Can I merge this?

PR #143, opened 2017-10-16
Added @*Title commands to specify titles for Chapters etc.

sebasguts:
which differ from the names given to the chapter/section/subsection

### DIFF
--- a/doc/Comments.xml
+++ b/doc/Comments.xml
@@ -139,10 +139,13 @@ text in your documentation, examples, mathematical chapters, etc..
 <Subsection Label="@Chapter">
 <Index Key="@Chapter"><C>@Chapter</C></Index>
 <Index Key="@ChapterLabel"><C>@ChapterLabel</C></Index>
+<Index Key="@ChapterTitle"><C>@ChapterTitle</C></Index>
 <Heading>@Chapter <A>name</A></Heading>
-Sets a chapter, all functions without separate info will be added to this chapter.
+Sets the active chapter, all subsequent functions which do not have an explicit chapter
+declared in their AutoDoc comment via <C>@ChapterInfo</C> will be added to this chapter.
 Also all text comments, i.e. lines that begin with #! without a command, and which do not
-follow after @description, will be added to the chapter as regular text.
+follow after <C>@Description</C>, will be added to the chapter as regular text. Additionally,
+the chapters label wil be set to <C>Chapter_</C><A>name</A>.
 
 Example:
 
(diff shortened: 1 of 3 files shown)


doc/Comments.xml
```diff
@@ -139,6 +139,7 @@ text in your documentation, examples, mathematical chapters, etc..
 <Subsection Label="@Chapter">
 <Index Key="@Chapter"><C>@Chapter</C></Index>
 <Index Key="@ChapterLabel"><C>@ChapterLabel</C></Index>
+<Index Key="@ChapterTitle"><C>@ChapterTitle</C></Index>
 <Heading>@Chapter <A>name</A></Heading>
 Sets a chapter, all functions without separate info will be added to this chapter.
```
fingolfin: While you are at it, perhaps also improve this text. How about: "Sets the active chapter. All subsequent functions which do not have an explicit chapter declared in their AutoDoc comment via [... err, how do we do it? Also via @Chapter?]  will be added to this chapter."

doc/Comments.xml
```diff
@@ -157,12 +158,14 @@ can be used to set the label of the chapter to <C>Chapter_</C><A>label</A>.
 
 Additionally, the chapter will be stored as <C>_Chapter_</C><A>label</A><C>.xml</C>.
 
-
+The <C>@ChapterTitle</C> <A>title</A> command
+can be used to set a heading for the chapter that is different com <A>name</A>
```
fingolfin: Typo: "com" -> "from" (I assume?)

Also, I am confused now: If I do this:
```
@Chapter Foo
@ChapterLabel Bar
@ChapterTitle Qux
```
Then I know `Bar` is used for the filename and the label. And `Qux` is used for the chapter heading. But what is `Foo` now used for?

doc/Comments.xml
```diff
@@ -173,6 +176,9 @@ Sets a section like chapter sets a chapter.
 
 The <C>@SectionLabel</C> <A>label</A> command
 can be used to set the label of the section to <C>Section_</C><A>label</A>.
+
+The <C>@SectionTitle</C> <A>title</A> command
+can be used to set a heading for the section that is different com <A>name</A>.
```
fingolfin: Typo "com"

doc/Comments.xml
```diff
@@ -202,6 +209,8 @@ Sets a subsection like chapter sets a chapter.
 The <C>@SubsectionLabel</C> <A>label</A> command
 can be used to set the label of the subsection to <C>Subsection_</C><A>label</A>.
 
+The <C>@SubsectionTitle</C> <A>title</A> command
+can be used to set a heading for the subsection that is different com <A>name</A>
```
fingolfin: Typo "com"

doc/Comments.xml
sebasguts: via `@ChapterInfo`. That's a good idea.

doc/Comments.xml
sebasguts: For opening the chapter again in the next file to put more sections in it.

doc/Comments.xml
```diff
@@ -139,10 +139,12 @@ text in your documentation, examples, mathematical chapters, etc..
 <Subsection Label="@Chapter">
 <Index Key="@Chapter"><C>@Chapter</C></Index>
 <Index Key="@ChapterLabel"><C>@ChapterLabel</C></Index>
+<Index Key="@ChapterTitle"><C>@ChapterTitle</C></Index>
 <Heading>@Chapter <A>name</A></Heading>
-Sets a chapter, all functions without separate info will be added to this chapter.
+Sets the active chapter, all All subsequent functions which do not have an explicit chapter
```
fingolfin: "all All" ?

doc/Comments.xml
```diff
@@ -157,14 +159,25 @@ can be used to set the label of the chapter to <C>Chapter_</C><A>label</A>.
...
+can be used to set a heading for the chapter that is different from <A>name</A>.
 
+If you use all three commands, i.e.,
+<Listing><![CDATA[
+#! @Chapter name
+#! @ChapterLabel label
+#! @ChapterTitle title
+]]></Listing>
+<C>title</C> is used for the headline, <C>label</C> for cross-referencing, and <C>name</C>
+for setting the same chapter as active chapter again.
```
fingolfin: So, if I only use `@Chapter` and `@ChapterTitle`, then what will the chapter label be?

fingolfin: No, that's fine. What I was getting at is: So essentially `@Chapter name` is equivalent to `@ChapterLabel name` plus `@ChapterTitle name` (of course it isn't quite the same, because you cannot omit it). 
If that's true, one would never want/need to use all three together. Also, the following two would be equivalent:
```
@Chapter label
@ChapterTitle title
```
vs
```
@Chapter title
@ChapterLabel label
```
Is that correct? If not, in which way do they differ?

If it *is* correct, then I wonder why we need `@ChapterTitle`...

sebasguts: They differ in a very particular point:
If you want to have content for the same chapter in different files, you have to reopen the chapter by the name you give it, i.e., in the first example one has to reopen the chapter with
```
@Chapter label
```
and in the second one with
```
@Chapter title
```
So, there actually is a reason to use all three of them together, I think.

sebasguts: The examples will by the way appear the same in the manual, but will for example produce different filenames.

fingolfin: I find this quite confusing -- why would one ever want to have three different values for these? (Not everything that *can* be done should be done).

However, since Chapter and ChapterLabel exists, we gotta support them for now... Let's just hope nobody will be caught by surprise by this ...

sebasguts:
Thank you :)

Issue #165, opened 2018-08-20
AutoDocWorksheet() does not succeed with absolute dir option


gwhitney:
As far as I can tell, at the moment AutoDocWorksheet() never succeeds if the options record contains a dir entry whose value is an IsDirectory() object consisting of an absolute path. Instead, it generates an error as follows:
```
gap> LoadPackage("AutoDoc");
─────────────────────────────────────────────────────────────────────────────
Loading  AutoDoc 2018.02.14 (Generate documentation from GAP source code)
by Sebastian Gutsche (http://wwwb.math.rwth-aachen.de/~gutsche/) and
   Max Horn (http://www.quendi.de/math).
Homepage: https://gap-packages.github.io/AutoDoc
─────────────────────────────────────────────────────────────────────────────
true
gap> tmpdir := DirectoryTemporary();;
gap> AutoDocWorksheet( "tst/worksheet.tst", rec( dir := tmpdir ) );
Generating documentation in Directory("/tmp/tm1UYs8b/")
Error, Variable: 'doc_dir_rel' must have an assigned value in
  tmp := Number( Filename( doc_dir_rel, "" ), function ( x )
        return x = '/';
    end )
 ; at [...]/gap-4.9.2/pkg/AutoDoc-2018.02.14/gap/Magic.gi:359 called from 
AutoDoc( "AutoDocWorksheet", arg[2] 
 ); at [...]/gap-4.9.2/pkg/AutoDoc-2018.02.14/gap/AutoDocMainFunction.gi:374 called from
<function "AutoDocWorksheet">( <arguments> )
```

I will momentarily submit a pull request with a proposed solution to this problem.

fingolfin:
A general problem with worksheets is that they are essentially undocumented (and I complained to @sebasguts about this before). So I have a hard time evaluating what is "right" or "expected" from them. From the manual, it's not even clear to me how to invoke `AutoDocWorksheet` correctly, and what the input should; I can't even tell if the invocation you give above (resp. in your pull request) makes sense.

I hope @sebasguts can chime in a bit...

gwhitney:
[...]

Issue #152, opened 2018-02-11
Add standard tests


alex-konovalov:
There is no `TestFile` component in the `PackageInfo.g` file. It is recommended to specify a short test (to run for no more than several minutes) which may be used to check that a package works as expected. For deposited packages, these tests are run regularly as a part of the standard GAP test suite. Please see (in GAP 4.9.0 or GAP development version) `?TestPackage`, `?Tests files for a GAP package` and also `?TestDirectory` for more information.

fingolfin:
Adding a meaningful test suite for AutoDoc is non-trivial, and `TestDirectory` is useless for this.

fingolfin:
See also #57 and #123

alex-konovalov:
`TestDirectory` is not the only way to test it of course. You can test something and analyse how did it go, and at the end do
```
if testresult then
  Print("#I  No errors detected while testing\n");
  QUIT_GAP(0);
else
  Print("#I  Errors detected while testing\n");
  QUIT_GAP(1);
fi;
```
like e.g. in https://github.com/gap-packages/qaos/blob/master/tst/testall.g

But indeed doing this in some meaningful way for Autodoc is tricky and requires some time to implement.

On a related note, the standard tests need not to be a test used in Travis. The same QaOS package now has a simple offline test specified as standard in PackageInfo.g and extended test for Travis. Perhaps #123 may also do something similar - just building the manual of AutoDoc itself, maybe couple of other packages, and collect code coverage statistics.

fingolfin:
Resolved by PR #183 and 7c90b2cee1c995797ac02a65c4c6aabe104769cb

Issue #57, opened 2014-04-08
Need more tests cases


fingolfin:
For quite some time, running AutoDoc() without the "autodoc" option was broken. We should have a test case to detect that. Likewise we should have test cases for other combinations of options being present / missing, to catch if this breaks again in the future.

In addition, it is bad if our changes break how existing packages are using AutoDoc. E.g. I am using AutoDoc() in the nq and polenta packages, but without the "autodoc" option, hence it is impossible to build their manual with various "bad" AutoDoc versions. 

Thus, I think it would make sense if we collected together all packages that are using AutoDoc in one way or another, and created a batch of tests for these and their documentation -- that is, we keep copies of their docs, and the tests rebuild them, and report a failure if there is any change in the output -- be it an error while running AutoDoc(), or be it a change in the generated HTML documentation.

PR #166, opened 2018-08-20
Handle absolute directories for doc_dir in AutoDoc()

gwhitney:
Allow AutoDoc() to succeed with a dir entry in the options record which
consists of an IsDirectory() object with an absolute path. There are two steps
to this change: First, some effort must be made to correctly compute
doc_dir_rel under such a circumstance. This commit uses the approach that if
the pkgdir is a prefix of the doc_dir, then doc_dir_rel is the remainder of
the doc_dir when the pkgdir is eliminated.

If doc_dir is absolute and pkgdir is not a prefix thereof, then there is no
sensible value for doc_dir_rel. Therefore, the second step of this change must
be to handle downstream the case that doc_dir_rel is not set. This commit
attempts to do so by presuming in such a case that there is no way to convert
the gapdoc.files to paths relative to the doc_dir, and instead converts them
to absolute paths. It may be the case that this strategy produces non-portable
documentation files, but at least MakeGAPDocDoc succeeds, and it seems to be
the only alternative in such case.

Finally, in my previous experience contributing to GAP there was a strong
preference that all changes include tests for the new behavior. As there was
no test harness for AutoDoc, this commit also adds a boilerplate tst/testall.g
and a single new test, tst/worksheet.tst, which invokes AutoDoc with an
absolutely specified doc_dir.

Resolves: #165.
In addition, the testing portion of this pull request seems to contribute some progress toward #152 and/or #57.

A minor point: I changed the Print() of the output directory to an Info() call so that I could suppress it by setting the InfoLevel to 0; since in the test, the output directory is a newly generated temp directory, the test would always fail if I could not suppress printing of the name of the temp directory.


gap/Magic.gi
```diff
@@ -205,7 +206,16 @@ function( arg )
...
 
     else
-        # TODO: doc_dir_rel = ... ?
+        # In this case, if doc_dir happens to lie below pkgdir, we want the
+        # doc_dir_rel to be the difference; if not we avoid binding doc_dir_rel
+        # and leave MakeGAPDocDoc to muddle through with absolute paths.
+        pkgdirstr := Filename(pkgdir, "");
+        docdirstr := Filename(doc_dir, "");
+        if Length(pkgdirstr) <= Length(docdirstr) and
+           docdirstr{[1..Length(pkgdirstr)]} = pkgdirstr then
```
fingolfin: more readable:
```gap
if StartsWith(docdirstr, pkgdirstr) then
```

gap/Magic.gi
```diff
@@ -351,14 +361,25 @@ function( arg )
...
-        # it does not contains '//') and relative.
-        # FIXME: this is an ugly hack, can't we do something better?
-        tmp := Number( Filename( doc_dir_rel, "" ), x -> x = '/' );
-        tmp := Concatenation( ListWithIdenticalEntries(tmp, "../") );
-        gapdoc.files := List( gapdoc.files, f -> Concatenation( tmp, f ) );
+        # If possible, convert the file paths in gapdoc.files, which are
+        # relative to the package directory, to paths which are relative to
+        # the doc directory.
+
+        if IsBound( doc_dir_rel) then
```
fingolfin: Please add a space after `doc_dir_rel`

tst/worksheet.tst
```diff
@@ -0,0 +1,62 @@
...
+##    Contributed by Glen Whitney, studioinfinity.org
+##
+## Licensed under the GPL 2 or later.
+##
+#############################################################################
+
+gap> START_TEST( "AutoDoc package: worksheet.tst" );
+gap> tmpdir := DirectoryTemporary();;
+gap> SetInfoLevel( InfoGAPDoc, 0 );
+gap> AutoDocWorksheet( "tst/worksheet.tst", rec( dir := tmpdir ) );
```
fingolfin: This seems to require that the tst file is run while the package dir is the current directory. But that's not necessarily true. The correct portable solution would normally be to use something like `Filename( DirectoriesPackageLibrary( "AutoDoc", "tst" ), "worksheet.tst")` instead -- but of course you want to test the relative filename here.

Perhaps one could add a `LoadPackage("io")` here and then use its `ChangeDirectoryCurrent` command to change to the "right" directory (and change back afterwards, to not risk breaking other tests).

tst/worksheet.tst
```diff
@@ -0,0 +1,62 @@
...
+gap> chap1 := Filename( Directory(tmpdir), "chap1.html" );;
+gap> IsReadableFile(chap1);
+true
+gap> ReadAll(InputTextFile(chap1));
```
fingolfin: I am not quite comfortable with this solution for testing the output. For once, it is prone to break when GAPDoc gets an update that changes the generated HTML. "Luckily", past experience tell us that GAPDoc updates are not exactly frequent.

But more importantly, this solution makes it very difficult to debug any test failures, because diff'ing the below text against the actually generated file is difficult. 

IMHO a better approach would be to store the expected solution as its own file, e.g. `tst/worksheet.expected/chap1.html`; then put the output of the test not into a random temp dir, but rather into `tst/worksheet.actual/` (which could be added to the `.gitignore`). Then the test here could luck roughly like this:
```
gap> ReadAll(InputTextFile(chap1_actual)) = ReadAll(InputTextFile(chap1_expected));
true
```
Then, if there is a test failure, I can just run `diff -u tst/worksheet.expected/chap1.html tst/worksheet.actual/chap1.html` to see what's going on.

tst/worksheet.tst
```diff
@@ -0,0 +1,62 @@
...
+rompt\">gap&gt;</span> <span class=\"GAPinput\">Size(S5);</span>\n120\n</pre><\
+/div>\n\n<p>And we wrap up with some dummy text</p>\n\n\n<div class=\"chlinkpr\
+evnextbot\">&nbsp;<a href=\"chap0.html\">[Top of Book]</a>&nbsp;  <a href=\"ch\
+ap0.html#contents\">[Contents]</a>&nbsp;  &nbsp;<a href=\"chap0.html\">[Previo\
+us Chapter]</a>&nbsp;  </div>\n\n\n<div class=\"chlinkbot\"><span class=\"chli\
+nk1\">Goto Chapter: </span><a href=\"chap0.html\">Top</a>  <a href=\"chap1.htm\
+l\">1</a>  </div>\n\n<hr />\n<p class=\"foot\">generated by <a href=\"http://w\
+ww.math.rwth-aachen.de/~Frank.Luebeck/GAPDoc\">GAPDoc2HTML</a></p>\n</body>\n<\
+/html>\n"
+gap> STOP_TEST( "plaintextmode.tst", 10000 );
```
fingolfin: Wrong filename.

tst/worksheet.tst
```diff
@@ -0,0 +1,62 @@
...
+#! @Date 2018/08/17
+#! @Chapter Test
+#! This is dummy text
+#! @BeginExampleSession
+#! gap> S5 := SymmetricGroup(5);
+#! Sym( [ 1 .. 5 ] )
+#! gap> Size(S5);
+#! 120
+#! @EndExampleSession
+#! And we wrap up with some dummy text
```
fingolfin: Actually, I'd prefer if the worksheet input was in a separate file, and not mixed up with the .tst file.

gwhitney:
Thank you for all of the excellent feedback. As soon as I can get to it, I am nearly certain tonight, I will create a new pull request with all of the suggested changes. My recollection is that the maintainers prefer a single-commit pull request as opposed to my adding a commit to this pull request, so I will simply make a new branch with the revised changes in a single commit, and submit a new pull request on that and close this one. (Unless there is a problem with that plan, in which case please let me know how to proceed. I remember that there are various operations for squashing commits and rebasing etc., but to be frank I don't use git on a daily basis and all of those operations seem very complex and confusing to me. So it's much easier for me to just checkout a clean master into a new branch, get the working directory into the desired state with the original changes plus the revisions outlined above, and then make a new pull request with a single commit  from that branch.)

tst/worksheet.tst
gwhitney: Indeed, in working with this approach I ran into exactly the same concerns, and in fact in the latest large pull request implements (among various things) almost exactly the changes you advocate; I switched to testing the XML rather than html, since AutoDoc's job, really, is just to generate the XML; and I added reference files and compared with them, rather than having the expected results in the .tst file.

However, there were three slight differences in my approach in that pull request as compared to what you are describing in this comment.

1) I still wrote the results into a tmpdir. I did not want to assume that the package directory was writable by the user id running the tests. (My mental model was one user on a multi-user machine who wants to run the tests against the installed common GAP installation; presumably in that case the user cannot write to the pkg directories of the common installation.) So my plan in the revised pull request will be to test if a directory of the form tst/worksheets/test1.actual can be created and is writable, and if so run the test there, but if not, fall back to a tmpdir.  If that doesn't seem like a good approach, let me know.

3) Rather than doing ReadAll on both and comparing the strings, I Exec()d a diff command to write its output to a file, and then did a ReadAll on the diffs. That way, if the test fails, the diffs are right there to see in the test report.

3) I deliberately put a small, one-line difference between the expected and actual, because to me, seeing that expected difference was more reassuring than completely silent output, which I worried could mask some issue where the entire operation went south and produced nothing. But I agree it is easier to deal with if the expected files don't have to be modified from AutoDoc's output when updating them. So I will switch to using the exact expected files, and use "diff -s -c" so that I have a message that the files are identical to check against for the success case. Again, if you would like any adjustments to that approach, just let me know.

gwhitney:
Superseded by #180, closing.

fingolfin:
Just FYI, there really is only a single command you need to learn to rebase/squash/etc. and that is `git rebase -i master`, i.e. an "interactive rebase". It is IMHO pretty intuitive to use. But your way is fine for us, too.
(And sorry for the long delays in reacting to you)

Issue #167, opened 2018-08-20
@BeginExampleSession/@EndExampleSession do not respect plain_text_mode


gwhitney:
If you save the attached .txt file as "tst/plaintextmode.tst" with respect to the current directory, and then execute the following GAP session:
```
gap> LoadPackage("AutoDoc");
─────────────────────────────────────────────────────────────────────────────
Loading  AutoDoc 2018.02.14 (Generate documentation from GAP source code)
by Sebastian Gutsche (http://wwwb.math.rwth-aachen.de/~gutsche/) and
   Max Horn (http://www.quendi.de/math).
Homepage: https://gap-packages.github.io/AutoDoc
─────────────────────────────────────────────────────────────────────────────
true
gap> SetInfoLevel(InfoGAPDoc, 0);
gap> AutoDocWorksheet( "tst/plaintextmode.tst", rec( dir := "tempdoc" ) );
Generating documentation in Directory("./tempdoc/")
gap> quit;
```
you indeed get a manual.pdf file (as also attached), and that manual does have an example block, but rather surprisingly, it is completely empty. On the other hand, if you prefix every line between the @BeginExampleSession and @EndExampleSession with `#!`, then you get the expected contents of the example block. This would appear to be a bug, that the state of plain_text_mode is not being respected in example sessions.

I will submit a pull request within a day with a proposed fix for this bug.
[manual.pdf](https://github.com/gap-packages/AutoDoc/files/2302086/manual.pdf)
[plaintextmode.txt](https://github.com/gap-packages/AutoDoc/files/2302087/plaintextmode.txt)

PR #168, opened 2018-08-20
Make @BeginExampleSession/@EndExampleSession respect plain_text_mode

gwhitney:
In order to uniformize the handling of `#!` prefixes in AutoDoc files,
gap/Parser.gi is changed to pass lines between @BeginExampleSession and
@EndExampleSession through unchanged when plain_text_mode is true.

Note this commit also supplies a test file tst/plaintextmode.tst; that file
will not have any effect unless/until a test harness such as the one supplied
with pull request #166 is put in place. But it does at least serve as a
self-documenting illustration of the intended behavior post this change, and
should otherwise be innocuous.

Resolves: #167


sebasguts:
I am not sure this is the entended behaviour.
`@ExampleSession` is meant to include examples that are prefixed with the `#!` comment. Yet the plain text mode does not require or remove `#!`. What is the intended use-case?

gwhitney:
Well, the documentation is pretty clear: "AutoDoc plain text files work exactly like AutoDoc comments, except that the #! is unnecessary at the beginning of a line which should be documented." I take from this that an AutoDoc plain text file should work just as if a #! is inserted at the beginning of every line and then AutoDoc operates as normally. It is very convenient for writing an expository section of a manual not associated with particular source code, with a very natural-looking .autodoc file. Consider a context like
```
@Section Computing the frob
The Baz package can also be used to compute the frob of a foo. To see how, consider the following example.
@BeginExampleSession
gap> f := MakeFoo();
<foo with 0 generators>
gap> g := Frobnicate(f);
<frob of f>
@EndExampleSession
It should be noted that since the frob may not be unique, g will be the lexicographically least frob satisfying the frob axioms for f.
```
It doesn't make any sense to me -- seems completely obvious that it was a bug -- that in a plain text file like this, one would have to introduce #! in front of just the lines between the @BeginExampleSession and @EndExampleSession, when they are not needed anywhere else, and indeed, would appear in the running text if they were present on the text lines. Why would such exceptional behavior for this one markup element be desirable?

My exact particular use case at the moment is that I am using my README.md file also as an introductory chapter of the manual, by having AutoDoc scan it. This actually works quite well with plain text mode, especially given that AutoDoc responds to a small but useful subset of Markdown notations. And I don't particularly want all of those #! appearing in the example GAP session on the front page of the code repository of my package -- they are confusing, and would not appear in the manual, so why should they be in my README?

You might say that whether or not each line of a transcript in a README is shown with a #! in front is a minor point.  Whether or not that's persuasive, it seems to me that the consistency argument of the operation of AutoDoc -- the #! prefix is needed and ignored in "normal mode" and unneeded (and treated as part of the text if present) in plain text mode -- is quite conclusive that #! should not be there in an example session in plain text mode. Plus, it makes composing a .autodoc file with examples much more straightforward: you execute a session, and then just copy and paste the transcript into your .autodoc file and surround it with @BeginExampleSession/@EndExampleSession, done. What could be more natural?

Thanks for your consideration.

tst/plaintextmode.tst
```diff
@@ -0,0 +1,73 @@
...
+@Chapter Test
+This is dummy text
+@BeginExampleSession
+gap> S5 := SymmetricGroup(5);
+Sym( [ 1 .. 5 ] )
+gap> Size(S5);
+120
+
+# This next command is present to prevent Test() from
+# being misled by the remaining AutoDoc markup.
```
fingolfin: That won't work, sorry. 

The correct solution is to put the actual worksheet into a separate file, say `tst/plaintextmode.worksheet`, and then read that, instead of trying to cram it all into the .tst file.

gwhitney: Well it certainly seemed to work, in that the test succeeded as expected. My motivation to merge both the testing code and the markup being tested in the same file was mostly that plaintextmode "should" work to interpolate plain text commands in any file AutoDoc() is processing, so it seemed perfectly reasonable that the test could process itself. A minor secondary motivation was economizing on the number of files in the test harness. But I am happy to re-do the pull request with separate worksheets and .tst file, and will do so ASAP, probably tonight.

gwhitney:
Superseded by #181, closing.

Issue #170, opened 2018-08-22
DeclareCategoryCollection causes strange behavior


gwhitney:
If you try to run AutoDoc on the following code snippet
```
#! @Chapter Test
#! @Section Some categories
#!  Intro text
DeclareCategory("MyThings", IsObject);
DeclareCategoryCollections("MyThings");
Now here is some text with a bunch of &!$%*!/ weird things in it. You'll
see that it ends up somewhere, shall we say, very unexpected.
```
it produces the following strange-looking error:
```
XML Parse Error: Line [XX] Character 86
Original file: ./crum/_Chapter_Test.xml, line number [YY].
-----------
  <Filt Arg="arg" Name="MyThings" Label="for ;Now here is some text with a bunch of &!$%*!/ weird things in\
 it. You'llsee that it ends up somewhere"/>
                                                                                     ^
-----------
no semicolon in entity reference
```
It appears that AutoDoc is confused and thinks there is another DeclareCategory on the DeclareCategoryCollections line, and then its argument parsing goes haywire. I will shortly post a pull request that adds handling for DeclareCategoryCollections.

fingolfin:
The problem here of course is that we are only looking for the string `DeclareCategory`, but we do not check what is before or after it. Hence we incorrectly "detect" `DeclareCategoryCollection` as `DeclareCategory`.
To fix this more generally, I think we should check that for any matched string, the next character is a non-identifier letter -- in fact, it should be a space of an opening parenthesis...

There are more "Declare*" operations sharing a prefix. Here are all `Declare*` defined in the GAP library; one may see lots of `*Kernel` variants; but also `DeclareSynonymAttr` (documented) and `DeclareCategoryFamily`, `DeclareAttributeSuppCT`, `DeclarePropertySuppCT` (all undocumented).
```
DeclareAttribute
DeclareAttributeKernel
DeclareAttributeSuppCT
DeclareAutoreadableVariables
DeclareCategory
DeclareCategoryCollections
DeclareCategoryFamily
DeclareCategoryKernel
DeclareConstructor
DeclareConstructorKernel
DeclareFilter
DeclareGlobalFunction
DeclareGlobalVariable
DeclareHandlingByNiceBasis
DeclareInfoClass
DeclareObsoleteSynonym
DeclareObsoleteSynonymAttr
DeclareOperation
DeclareOperationKernel
DeclareProperty
DeclarePropertyKernel
DeclarePropertySuppCT
DeclareRepresentation
[...]

PR #171, opened 2018-08-23
 Handle documentation of DeclareCategoryCollection declarations. 

gwhitney:
Note that AutoDoc here duplicates logic in `gap/lib/coll.gd` to determine the
actual name of the symbol being defined by the call (since that symbol is
computed on the fly by the declaration and does not appear verbatim anywhere
in the call). This is unfortunate, but I could not see any way to call the
relevant code in the standard library, as the name computation of the
collections category is not split out into a separately executable function.

Also, modify the `worksheet.tst` to include an example of
`DeclareCategoryCollection` to make sure the new capacity is tested.

Note that this pull request is on top of the changes made for pull request #166, because otherwise I wasn't sure how to ensure that this pull request included test for the new facility; I just needed to be on top of a test harness.

Resolves: #170

### DIFF
--- a/gap/Parser.gi
+++ b/gap/Parser.gi
@@ -63,7 +63,15 @@ InstallGlobalFunction( AutoDoc_Type_Of_Item,
   function( current_item, type, default_chapter_data )
     local item_rec, entries, has_filters, ret_val;
     item_rec := current_item;
-    if PositionSublist( type, "DeclareCategory" ) <> fail then
+    if PositionSublist( type, "DeclareCategoryCollections") <> fail then
+        entries := [ "Filt", "categories" ];
+        ret_val := "<C>true</C> or <C>false</C>";
+        has_filters := "No";
+        if not IsBound( item_rec!.arguments ) then
+            item_rec!.arguments := "obj";
+        fi;
+        item_rec!.coll_suffix := true;
+    elif PositionSublist( type, "DeclareCategory" ) <> fail then
         entries := [ "Filt", "categories" ];
         ret_val := "<C>true</C> or <C>false</C>";
         has_filters := 1;
@@ -229,7 +237,7 @@ InstallGlobalFunction( AutoDoc_Parser_ReadFiles,
                 return false;
             fi;
             current_line := current_line{ [ position_parentesis + 1 .. Length( current_line ) ] };
-            ## Not the funny part begins:
+            ## Now the funny part begins:
             ## try fetching the name:
             ## Assuming the name is in the same line as its
             while PositionSublist( current_line, "," ) = fail and PositionSublist( current_line, ");" ) = fail do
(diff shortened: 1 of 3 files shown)


fingolfin:
This PR seems sensible in general, but will have to wait till PR #180 is resolved and merged (and then this PR will need to be rebased, I am afraid)

fingolfin:
Oh, and once this is rebased, I am all for merging it, the change is useful.

fingolfin:
I've dropped the first (unrelated) commit and rebased this (which mostly meant updating the test). I think we could merge it now.

gap/Parser.gi
```diff
@@ -229,7 +237,7 @@ InstallGlobalFunction( AutoDoc_Parser_ReadFiles,
                 return false;
             fi;
             current_line := current_line{ [ position_parentesis + 1 .. Length( current_line ) ] };
-            ## Not the funny part begins:
+            ## Now the funny part begins:
```
sebasguts: Hehe, nice.

gap/Parser.gi
```diff
@@ -238,6 +246,19 @@ InstallGlobalFunction( AutoDoc_Parser_ReadFiles,
...
+                    current_item!.name{[1..Length(current_item!.name)-6]};
+                fi;
+                if EndsWith(current_item!.name, "Coll") then
+                    current_item!.coll_suffix := "Coll";
+                else
+                    current_item!.coll_suffix := "Collection";
+                fi;
+                current_item!.name := Concatenation(current_item!.name,
+                                                    current_item!.coll_suffix);
+            fi;
```
sebasguts: Can you please explain what this `if` does?

To me it seems like you check if the item ends with `Collection`, you do nothing. If it ends with `Coll`, you add another `Coll` to the name?

Specially, in the docs below it seems like `MyThingsColl` gets a manual entry `MyThingsCollColl`, which seems wrong . . .

fingolfin: As explained in the PR description and the commit message, this imitates the behavior of `DeclareCategoryCollections`. I recommend looking at the code for `DeclareCategoryCollections` (and the comment before it) in the GAP library. Let me know if there are still questions after that :-).

That said, I think there should be a comment here which explains that; I can add one later.

sebasguts: I see . . .

Okay, sorry for the confusion.

fingolfin: No confusion, nothing to be sorry about: you raised a valid point. I've now added a comment, which hopefully makes this code easier to understand for future generations (e.g. us, two years down the road ;-).

fingolfin: I will also submit a PR to GAP to actually document `DeclareCategoryCollections`, which lots of packages use...

In the meantime, feel free to approve this PR ;-)

Issue #90, opened 2015-04-10
Headings for groups


oysteins:
Can I specify the heading for a group in AutoDoc?

I have observed that if I create a group in AutoDoc using @BeginGroup and @EndGroup, then in the generated GAPDoc code, there is a single ManSection element containing all functions in the group, but with no Heading element.  GAPDoc then uses the name of the first function in the ManSection as heading for the section.  This is not always what I want.  My question is thus: Can I somehow get AutoDoc to insert a Heading element in the ManSection it generates for a group, containing some arbitrary text specified by me?

fingolfin:
@sebasguts Can you answer this?

sebasguts:
Currently this is not possible, but I noticed that GAPDoc is able to support this. Would it be a good idea to have a ```@GroupHead``` command for this? Maybe one does not want to have the name of the group inserted by default?

oysteins:
Yes, a `@GroupHead` command (which takes some text as argument and generates a `Heading` element containing this text) would resolve the issue.  Sorry for the very late reply.

PR #172, opened 2018-08-23
Add @GroupTitle and @GroupInitialArguments commands

gwhitney:
Previously, there was no way to modify the subsection heading of a Group, but
the documentation even stated "Note that it is currently not possible to give
a header to the Group in the manual..." Hence, the addition of a @GroupTitle
command to allow exactly that.

In addition, the functions in a Group often share an initial set of arguments,
and so this adds a @GroupInitialArguments command to allow such arguments to
be defined once for the declarations in a @BeginGroup/@EndGroup block.

Both commands are added to the documentation, with examples in the Grouping
section.

Resolves: #90. However, that issue called for the command to be called @GroupHead. I was not aware of the issue when writing this code; I chose @GroupTitle by analogy with @SectionTitle and @SubsectionTitle (as a Group is just a special kind of Subsection). Let me know if you would like a version of the pull request with the command changed to @GroupHead, or with just the group title command and not the inital arguments command, or both, and I would be happy to supply it.

### DIFF
--- a/doc/Comments.xml
+++ b/doc/Comments.xml
@@ -302,6 +302,25 @@ DeclareOperation( "GroupedOperation",
 ]]></Listing>
 </Subsection>
 
+<Subsection Label="@GroupTitle">
+<Index Key="@GroupTitle"><C>@GroupTitle</C></Index>
+<Heading>@GroupTitle <A>title</A></Heading>
+Sets the subsection heading for the current group to <A>title</A>. In the
+absence of any <C>@GroupTitle</C> command, the heading will be the name of the
+first entry in the group. See <Ref Sect="Groups" /> for more information.
+</Subsection>
+
+<Subsection Label="@GroupInitialArguments">
+<Index Key="@GroupInitialArgumentss"><C>@GroupInitialArguments</C></Index>
+<Heading>@GroupInitialArguments <A>args</A></Heading>
+Sets the common initial arguments for the entries in the current group to be
+<A>args</A>. The argument list for any function subsequently added to the
+group without any args specified will be <A>args</A>, and <A>args</A> will be
+prepended to any arguments specified for a particular function. The setting is
+cleared by an <C>@EndGroup</C> command. See <Ref Sect="Groups" /> for more
+information.
+</Subsection>
+
 <Subsection Label="@Level">
 <Index Key="@Level"><C>@Level</C></Index>
 <Heading>@Level <A>lvl</A></Heading>
@@ -556,8 +575,9 @@ a function, operation, etc., it is possible to group them into suitable chunks.
 This can be particularly useful if there are several definitions of an operation
 with several different argument types, all doing more or less the same to the arguments.
 Then their manual items can be grouped, sharing the same description and return type information.
-Note that it is currently not possible to give a header to the Group in the manual,
-but the generated ManItem heading of the first entry will be used.
+You can give a heading to the Group in the manual with the <C>@GroupTitle</C>
+command; if that is not supplied, then the generated ManItem heading of the
+first entry will be used as the heading.
 <P/>
 
 Note that group names are globally unique throughout the whole manual.
@@ -570,9 +590,19 @@ same group name, in different places, and these all will refer to the same group
 Moreover, this means that you can add items to a group via the <C>@Group</C> command
 in the &AutoDoc; comment of an arbitrary declaration, at any time.
 
+<P/>
+Since often the functions in a Group share common initial arguments, you can
+set a list of common arguments to be prepended to all declarations within a
+given <C>@BeginGroup/@EndGroup</C> pair with the <C>@GroupInitialArguments</C>
+command.
+
+<P/>
+
 The following code
 <Listing><![CDATA[
 #! @BeginGroup Group1
+#! @GroupTitle A family of operations
+#! @GroupInitialArguments order
 
 #! @Description
 #!  First sentence.
@@ -580,6 +610,7 @@ DeclareOperation( "FirstOperation", [ IsInt ] );
 
 #! @Description
 #!  Second sentence.
+#! @Arguments ambient_group 
 DeclareOperation( "SecondOperation", [ IsInt, IsGroup ] );
 
 #! @EndGroup
@@ -595,8 +626,9 @@ KeyDependentOperation( "ThirdOperation", IsGroup, IsInt, "prime );
 produces the following:
 
 <ManSection Label="Group1">
-  <Oper Arg="arg" Name="FirstOperation" Label="for IsInt"/>
-  <Oper Arg="arg1,arg2" Name="SecondOperation" Label="for IsInt, IsGroup"/>
+  <Heading>A family of operations</Heading>
+  <Oper Arg="order" Name="FirstOperation" Label="for IsInt"/>
+  <Oper Arg="order,ambient_group" Name="SecondOperation" Label="for IsInt, IsGroup"/>
   <Oper Arg="arg1,arg2" Name="ThirdOperation" Label="for IsGroup, IsInt"/>
  <Returns></Returns>
  <Description>
--- a/gap/DocumentationTree.gi
+++ b/gap/DocumentationTree.gi
@@ -623,10 +623,15 @@ end );
 ##
 InstallMethod( WriteDocumentation, [ IsTreeForDocumentationNodeForGroupRep, IsStream ],
   function( node, filestream )
+    local head;
     if node!.level > ValueOption( "level_value" ) then
         return;
     fi;
-    AutoDoc_WriteDocEntry( filestream, node!.content );
+    head := fail;
+    if IsBound( node!.title_string ) then
+        head := node!.title_string;
+    fi;
+    AutoDoc_WriteDocEntry( filestream, node!.content, head );
 end );
 
 ##
--- a/gap/Parser.gi
+++ b/gap/Parser.gi
@@ -77,7 +77,11 @@ InstallGlobalFunction( AutoDoc_Type_Of_Item,
         entries := [ "Func", "global_functions" ];
         has_filters := "No";
         if not IsBound( item_rec!.arguments ) then
-            item_rec!.arguments := "arg";
+            if IsBound( item_rec!.initial_args ) then
+                item_rec!.arguments := item_rec!.initial_args;
+            else
+                item_rec!.arguments := "arg";
+            fi;
         fi;
     elif PositionSublist( type, "DeclareGlobalVariable" ) <> fail then
         entries := [ "Var", "global_variables" ];
@@ -116,7 +120,7 @@ InstallGlobalFunction( AutoDoc_Parser_ReadFiles,
   function( filename_list, tree, default_chapter_data )
     local current_item, flush_and_recover, chapter_info, current_string_list,
           Scan_for_Declaration_part, flush_and_prepare_for_item, current_line, filestream,
-          level_scope, scope_group, read_example, command_function_record, autodoc_read_line,
+          level_scope, scope_group, scope_initial_args, read_example, command_function_record, autodoc_read_line,
           current_command, was_declaration, filename, system_scope, groupnumber, chunk_list, rest_of_file_skipped,
           context_stack, new_man_item, add_man_item, Reset, read_code, title_item, title_item_list, plain_text_mode,
           current_line_unedited,
@@ -158,6 +162,9 @@ InstallGlobalFunction( AutoDoc_Parser_ReadFiles,
         fi;
         if IsBound( scope_group ) then
             SetGroupName( man_item, scope_group );
+        fi;
+        if IsBound( scope_initial_args ) then
+            man_item!.initial_args := scope_initial_args;
         fi;
         man_item!.chapter_info := ShallowCopy( chapter_info );
         man_item!.tester_names := fail;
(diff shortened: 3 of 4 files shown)


gap/ToolFunctions.gi
```diff
@@ -25,7 +25,7 @@ end );
 
 ##
 InstallGlobalFunction( AutoDoc_WriteDocEntry,
-  function( filestream, list_of_records )
+  function( filestream, list_of_records, maybe_heading... )
```
fingolfin: I don't like having `maybe_heading` an optional argument. Just change `maybe_heading...` to `heading`.

gap/ToolFunctions.gi
```diff
@@ -78,7 +78,15 @@ InstallGlobalFunction( AutoDoc_WriteDocEntry,
...
     AppendTo( filestream, ">\n" );
 
-    # Function heades
+    # Next possibly the heading for the entry
+    if Length( maybe_heading ) > 0 then
+        maybe_heading := maybe_heading[ 1 ];
+        if IsString( maybe_heading ) then
+            AppendTo( filestream, "<Heading>", maybe_heading, "</Heading>\n" );
+        fi;
+    fi;
```
fingolfin: With the change to `heading`, this could be replaced by
```gap
if IsString( heading ) then
  AppendTo( filestream, "<Heading>", heading, "</Heading>\n" );
fi;
```

gap/DocumentationTree.gi
```diff
@@ -623,10 +623,15 @@ end );
...
+    local head;
     if node!.level > ValueOption( "level_value" ) then
         return;
     fi;
-    AutoDoc_WriteDocEntry( filestream, node!.content );
+    head := fail;
+    if IsBound( node!.title_string ) then
+        head := node!.title_string;
+    fi;
+    AutoDoc_WriteDocEntry( filestream, node!.content, head );
```
fingolfin: Please rename `head` to `heading`.

doc/Comments.xml
```diff
@@ -302,6 +302,25 @@ DeclareOperation( "GroupedOperation",
 ]]></Listing>
 </Subsection>
 
+<Subsection Label="@GroupTitle">
+<Index Key="@GroupTitle"><C>@GroupTitle</C></Index>
+<Heading>@GroupTitle <A>title</A></Heading>
+Sets the subsection heading for the current group to <A>title</A>. In the
+absence of any <C>@GroupTitle</C> command, the heading will be the name of the
+first entry in the group. See <Ref Sect="Groups" /> for more information.
```
fingolfin: `@GroupTitle` seems like a sensible addition to me.

doc/Comments.xml
```diff
@@ -302,6 +302,25 @@ DeclareOperation( "GroupedOperation",
...
+</Subsection>
+
+<Subsection Label="@GroupInitialArguments">
+<Index Key="@GroupInitialArgumentss"><C>@GroupInitialArguments</C></Index>
```
fingolfin: Typo: `GroupInitialArgumentss` with double s

doc/Comments.xml
```diff
@@ -302,6 +302,25 @@ DeclareOperation( "GroupedOperation",
...
+<Heading>@GroupInitialArguments <A>args</A></Heading>
+Sets the common initial arguments for the entries in the current group to be
+<A>args</A>. The argument list for any function subsequently added to the
+group without any args specified will be <A>args</A>, and <A>args</A> will be
+prepended to any arguments specified for a particular function. The setting is
+cleared by an <C>@EndGroup</C> command. See <Ref Sect="Groups" /> for more
+information.
```
fingolfin: I don't like `@GroupInitialArguments` at all. It makes information non-local: with it, when reading a function's AutoDoc comment, I must be aware of whether it is inside a group, and whether a `@GroupInitialArguments` was specified or not.

This does not seem to justify the (IMHO very minor, if at all) "savings" provided by it. It's not that difficult to repeat the string `order, ` two or three times (it won't be more in general).

doc/Comments.xml
```diff
@@ -556,8 +575,9 @@ a function, operation, etc., it is possible to group them into suitable chunks.
 This can be particularly useful if there are several definitions of an operation
 with several different argument types, all doing more or less the same to the arguments.
 Then their manual items can be grouped, sharing the same description and return type information.
-Note that it is currently not possible to give a header to the Group in the manual,
-but the generated ManItem heading of the first entry will be used.
+You can give a heading to the Group in the manual with the <C>@GroupTitle</C>
```
fingolfin: Group -> group (yes, this was misspelled before.

doc/Comments.xml
```diff
@@ -556,8 +575,9 @@ a function, operation, etc., it is possible to group them into suitable chunks.
...
+command; if that is not supplied, then the generated ManItem heading of the
```
fingolfin: `ManItem` is also bad here, and should not be used -- but of course it was here before, so I do not expect you to fix it (rather, I expect me or @sebasguts to fix it)

fingolfin:
Oh and of course a test case would be great for this -- for that, you'll have to rebase on the latest master. It shouldn't be that hard:
```bash
git checkout master
git pull
git checkout NAME_OF_THIS_BRANCH
git rebase master
```
If anything goes wrong, you can always `git rebase --abort`

fingolfin:
Just to clarify re: tests: first off, this changes here will of course affect the `dogfood.tst`, so that's one reason we need a rebase.

Secondly, of course `@GroupTitle` will get tested because it is used in the manual, hence `dogfood.tst` will test it. But now we lack a test for `@Group` *without* a @GroupName`. So I think we really want a set of tests (say via worksheets in `tst`) that test every single feature of AutoDoc, in various combinations.

But of course none of that is relevant for *this* PR -- I only mention it as a reminder to self for future work.

doc/Comments.xml
sebasguts: Why do you think this is bad? `ManItem` is a special kind of subsection in GAPDoc, which does not need to have a heading, but will create one. What would you suggest?

sebasguts:
As @fingolfin pointed out, `@GroupTitle` is very useful and was probably long overdue, thank you very much.
But I have also concerns the initial arguments are very useful. It makes reading the docs in the source code harder (that's whereI read documentations most of the time and I would guess many people do the same), and it serves no immediate benefit.
In some sense one could think about an API to extend the AutoDoc Syntax sensibly, but I am not sure if that is so easy.

doc/Comments.xml
fingolfin: First off, it should at the very least be typeset with `<C>` or so -- better would be a Ref to the relevant GAPDoc manual entry, though.

But really, why is a technical term from GAPDoc used to explain things here? In order to understand this, now one needs to be familiar with both AutoDoc and GAPDoc, *and* how the conversion from one to the other works exactly.

It would be much better if the AutoDoc explanation was self-contained. Perhaps in the future, we do not just output GAPDoc, but also some other format?

At the same time, it can be useful to document how certain AutoDoc constructs translate to GAPDoc. But IMHO this should be separate (say, in a separate paragraph). And then for more (ideally all) AutoDoc, not just a random (?) aubset

sebasguts: Yep, you are right, it is probably bad to refer to GAPDoc internals (probably out of laziness here). Anyway, as you stated, I should probably fix this myself, so this should not block this PR.

gwhitney:
My apologies for being a bit slow in responding to all of your great feedback/requests for changes to PRs; the semester has started, and so demands of teaching have given me less time for these matters. But they are high on my list and I will tend to all of them as soon as I possibly can.  But I wanted quickly to show my exact use case for the initial arguments command for the actual package I am working on, which is an evolution of the RIG almost-but-not-quite official package for racks in gap, trying to bring it up to generally accepted GAP standards in terms of data structures, interfaces, and documentation, while improving speed/correctness of several of its algorithms.

Anyhow, in the documentation, I use Groups extensively for collections of very similar functions that differ only in the algebraic category they concern, e.g:

```
#! @BeginGroup basic_constructors
#! @GroupTitle Basic constructors (from generators)
#! @Description These are the fundamental constructors of these
#!   categories. They produce the closure of the specified <A>generators</A>,
#!   which are considered to be of the given <A>family</A>, under both the
#!   binary operation of the magma, which is always considered to be * in
#!   &RAQ;, and the quotient on the specified side. (If
#!   <A>family</A> is omitted, these functions attempt to infer it from the
#!   <A>generators</A>; if there are no <A>generators</A> then the
#!   <A>family</A> must be specified, and note that the resulting magma will
#!   be empty.) The resulting magma must satisfy the defining
#!   characteristics of the respective category: for the quasigroups, all
#!   quotients on the specified side must exist; racks must also satisfy the
#!   appropriate self-distributive law; and quandles must also have every
#!   element idempotent.
#! @Returns a magma of the designated category
#! @GroupInitialArguments [family], [generators]
DeclareGlobalFunction("LeftQuasigroup");
DeclareGlobalFunction("LeftRack");
DeclareGlobalFunction("LeftQuandle");
DeclareGlobalFunction("RightQuasigroup");
DeclareGlobalFunction("RightRack");
DeclareGlobalFunction("RightQuandle");
#! @EndGroup
```
To me, specifying `@Arguments` repeatedly and identically for each of these functions is not only tedious, it violates the DRY (Don't Repeat Yourself) principle fairly egregiously: all of these functions are naturally going to have the same arguments, and if it turns out that in the future one needs to add, say, a "relations" argument giving additional relations on the generators (yes, that's a made-up, unlikely eventuality, but who knows what will come up) it will surely be added to all of them. In fact, when I first started using AutoDoc, I was astonished that `@Arguments` didn't apply to the entire group the way `@Returns` does. Then, of course, I encountered the "other" use case for Group, which is a bunch of functions that produce nearly identical output but from _different_ inputs.  (And moreover, note there have been comments here that in some cases people want _different_ `@Returns` for different members of a group.) But for this use case, which I think is an important use of Groups, I essentially "need" some way of specifying the arguments in a single place. 

I don't understand the "non-locality" argument against a common arguments-setter, in light of the fact that is exactly how `@Returns` currently behaves with groups.

In some sense, at issue is the asymmetry between `@Arguments` and `@Returns` for Groups -- given the different use cases, there is not (to me) any strong argument for that asymmetry.

I considered only implementing a  `@GroupArguments` command that would set and mandate the arguments for all members of a group, but as I was implementing it, it was just as easy to make the initial arguments version, which was more flexible -- I could imagine a use case in which one of the variants did have an additional argument special to it, so  it seemed I might as well. There's nothing compelling the use of `@GroupInitialArguments`, so anyone concerned about the non-locality (or with other concerns) could simply avoid it and write out all of the arguments.

But I would argue that the documentation for the rack/quandle package has a very strong need for a command that sets the arguments once for the whole group, and that the non-parallelism of `@Arguments` and `@Returns` is unmotivated. 

Please provide an acceptable design/spec for a command that allows arguments to be set once for an entire group. It's quite important to my project, and so I am happy to implement whatever spec you find reasonable.  Two possibilities other than what I proposed in this pull request (but I am open to anything you decide would be best):

- `@GroupArguments` sets the arguments once and for all for a group, and mixing of `@Arguments` and `@GroupArguments` is an error
- `@GroupArgumentsDefault` sets the arguments for any member of a group that does not have its own `@Arguments` command; any member that does ignores the `@GroupArgumentsDefault`.

If it would improve acceptability, in any design you pick I would be happy to make the `@Returns` design completely parallel: `@GroupInitialReturns` specifies a prefix of the returns text for the group, which individual `@Returns` can add to, or `@GroupReturns` acts basically like `@Returns` does now and it is illegal to mix it with individual `@Returns`, or `@GroupReturnsDefault` sets an overall returns that can be overridden in specific members of the group with `@Returns`. Or I am happy to leave `@Returns` entirely alone, as I did in the original PR.

Please help me find a path to a common argument declaration for a whole group; I use it numerous times in the the documentation for this racks package, and I think it makes the documentation in the source code much _more_ readable. Thank you.

gap/DocumentationTree.gi
```diff
@@ -623,10 +623,15 @@ end );
 ##
 InstallMethod( WriteDocumentation, [ IsTreeForDocumentationNodeForGroupRep, IsStream ],
   function( node, filestream )
+    local head;
```
fingolfin:
```suggestion
    local heading;
```

gap/DocumentationTree.gi
```diff
@@ -623,10 +623,15 @@ end );
...
     if node!.level > ValueOption( "level_value" ) then
         return;
     fi;
-    AutoDoc_WriteDocEntry( filestream, node!.content );
+    head := fail;
```
fingolfin:
```suggestion
    heading := fail;
```

gap/DocumentationTree.gi
```diff
@@ -623,10 +623,15 @@ end );
...
+    if IsBound( node!.title_string ) then
+        head := node!.title_string;
```
fingolfin:
```suggestion
        heading := node!.title_string;
```

gap/DocumentationTree.gi
fingolfin:
```suggestion
    AutoDoc_WriteDocEntry( filestream, node!.content, heading );
```

gap/Parser.gi
```diff
@@ -593,6 +617,28 @@ InstallGlobalFunction( AutoDoc_Parser_ReadFiles,
...
+                ErrorWithPos( "found @GroupTitle with no Group set" );
+            fi;
+            group_name := GroupName( current_item );
+            chap_info := fail;
+            if HasChapterInfo( current_item ) then
+                chap_info := ChapterInfo( current_item );
+            elif IsBound( current_item!.chapter_info ) then
+                chap_info := current_item!.chapter_info;
+            fi;
+            if (chap_info = fail or chap_info = [ ]) then
```
fingolfin:
```suggestion
            if chap_info = fail or chap_info = [ ] then
```

gap/ToolFunctions.gi
fingolfin: To be clearer: I suggest to change this to this:
```suggestion
  function( filestream, list_of_records, heading )
```
Then callers should either pass a string for `heading`, or something else, like `fail`, to indicate no heading is desired. That'll make the code a bit simpler, I'd say.

Issue #163, opened 2018-08-20
Minor omissions/imprecisions in AutoDoc() function documentation


gwhitney:
1. The argument list for AutoDoc() in the manual implies that if the package parameter is omitted, the optrec must be as well, but that's not the case.

2. The list of tasks AutoDoc() can perform omits the creation of maketest.g.

3. The package parameter serves to determine both the package directory in which AutoDoc() will operate and the metadata for the package being documented. It would be useful to clarify this point and explain exactly how those two pieces of information are determined in each of the three cases for the parameter (a string, an IsDirectory, or omitted). 

4. It would also be nice for completeness to say how those two things are determined in the worksheet case, since the documentation for AutoDocWorksheet() primarily just refers to the AutoDoc() documentation.

5. The documentation for the dir entry of the options record hints at the fact that a string value for that option must always be relative to the package directory. It would be useful to emphasize/clarify this point, and also to document that this is not only where documentation files like GapDoc XML will be found, but also where all files will be generated.

I will momentarily submit a pull request with suggested resolutions of these (admittedly minor) items. The motivation for filing this issue and pull request at all is that I also believe there is a bug with using an absolute dir (in the form of an IsDirectory object) in the options record at all, and I plan shortly to submit an issue for that bug and a pull request resolving it, but it seemed necessary to document the apparently intended behavior of that parameter and related issues first. If I've misunderstood the intended behavior of AutoDoc(), my apologies and of course that should be taken into account in evaluating the subsequent issue.

Issue #155, opened 2018-02-13
Using maketest.g in package standard tests


alex-konovalov:
How would you combine `maketest.g` with existing `tst/testall.g` using `TestDirectory` setup?

fingolfin:
Good question. I'd hope @sebasguts can answer, also #114 and #117

Issue #117, opened 2016-03-07
Replace `maketest` option by a new AutoDocRunExamples() function


fingolfin:
Right now, the `maktest` function outputs a `maketest.g` script which can look like this (wrapped for readability):

```
## This file is automatically generated by AutoDoc. Changes will be
## discarded by the next call of the AutoDoc method.
## 
## 
AUTODOC_file_scan_list := [ "../ListOfDocFiles.g", "../PackageInfo.g",
"../examples/ExtExt.g", "../examples/Grothendieck-SpectralCosequence.g",
"../examples/Grothendieck.g",
"../examples/Hom(Hom(-,Z128),Z16)_On_Seq.g",
"../examples/LTensorProduct_Z.g", "../examples/Purity.g",
"../examples/RHom_Z.g", "../examples/Saturate.g",
"../examples/SpectralCosequence.g", "../examples/SpectralSequence.g",
"../examples/TorExt.g", "../examples/TorExt_Grothendieck.g",
"../examples/TorExt_via_Hom(P,P).g", "../examples/Triangle.g",
"../examples/WhySpectral.g", "../examples/doc/HomalgElement_in.g",
"../examples/homalg.g", "../examples/qiCounterExample.g",
"../gap/BasicFunctors.gd", "../gap/BasicFunctors.gi",
"../gap/ChainMorphisms.gd", "../gap/ChainMorphisms.gi",
"../gap/Complexes.gd", "../gap/Complexes.gi", "../gap/Filtrations.gd",
"../gap/Filtrations.gi", "../gap/HomalgBicomplex.gd",
"../gap/HomalgBicomplex.gi", "../gap/HomalgBigradedObject.gd",
"../gap/HomalgBigradedObject.gi", "../gap/HomalgCategory.gd",
"../gap/HomalgCategory.gi", "../gap/HomalgChainMorphism.gd",
"../gap/HomalgChainMorphism.gi", "../gap/HomalgComplex.gd",
"../gap/HomalgComplex.gi", "../gap/HomalgDiagram.gd",
"../gap/HomalgDiagram.gi", "../gap/HomalgElement.gd",
"../gap/HomalgElement.gi", "../gap/HomalgFiltration.gd",
"../gap/HomalgFiltration.gi", "../gap/HomalgFunctor.gd",
"../gap/HomalgFunctor.gi", "../gap/HomalgMorphism.gd",
"../gap/HomalgMorphism.gi", "../gap/HomalgObject.gd",
[...]

Issue #114, opened 2016-03-04
Alternative to maketest?


fingolfin:
So @alex-konovalov pointed out to me that his packages (e.g. wedderga) use the following approach towards manual examples: Similar our `maketest` code, he implemented a function which extracts manual examples, and makes them easy to run.
But his approach is to turn the manual examples into `tst/*.tst` files, whose names is based on the chapters they are from. The tests then are run like any other test, assuming your package runs all `tst/*.tst` files anyway. Here is the function he uses for that:

``` gap
ExtractMyManualExamples:=function( pkgname, main, files )
local path, tst, i, s, name, output, ch, a;
path:=Directory( 
        Concatenation(PackageInfo(pkgname)[1].InstallationPath, "/doc") );
Print("Extracting manual examples for ", pkgname, " package ...\n" );
tst:=ExtractExamples( path, main, files, "Chapter" );
Print(Length(tst), " chapters detected\n");
for i in [ 1 .. Length(tst) ] do 
  Print( "Chapter ", i, " : \c" );
  if Length( tst[i] ) > 0 then
    s := String(i);
    if Length(s)=1 then 
      # works for <100 chapters
      s:=Concatenation("0",s); 
    fi;
    name := Filename( 
              Directory( 
                Concatenation( PackageInfo(pkgname)[1].InstallationPath, 
                               "/tst" ) ), 
                Concatenation( LowercaseString(pkgname), s, ".tst" ) );
    output := OutputTextFile( name, false ); # to empty the file first
    SetPrintFormattingStatus( output, false ); # to avoid line breaks
    ch := tst[i];
    AppendTo(output, "# ", pkgname, ", chapter ",i,"\n");
    for a in ch do
      AppendTo(output, "\n# ",a[2], a[1]);
    od;
    Print("extracted ", Length(ch), " examples \n");
  else
    Print("no examples \n" );    
  fi;  
od;
end;
```

It would be easy to integrate this into `AutoDoc`, with some slight changes (e.g. instead of a package name, it should take a package info record).

I wonder whether we should do something similar -- either in addition to `maketest`, or as a replacement? I wonder what the pros and cons of each approach are? Perhaps creating one tst file per chapter makes it easier to track down where a problem sits, exactly?

@sebasguts BTW, can you point me to any projects using the maketest feature, so that I can try it and learn how ti works, exactly?

alex-konovalov:
[...]

PR #173, opened 2018-08-24
Repair minor omissions/imprecisions in AutoDoc() function doc

gwhitney:
Specifically:

1. Modify the argument list to indicate that the two parameters are each
independently optional.

2. Add creating a maketest.g to the list of tasks AutoDoc() can perform.

3. Clarify that the package parameter determines both the package directory in
which AutoDoc will operate and the metadata for the package being documented,
and describe more precisely how it does so.

4. Clarify how the package directory and metadata work in the worksheet case.

5. Clarify the semantics of the dir entry in the options record.

Resolves: #163 (note this replaces a former pull request to resolve that issue which became "dirty" with additional commits).

### DIFF
--- a/gap/Magic.gd
+++ b/gap/Magic.gd
@@ -11,7 +11,7 @@
 
 #! @Description
 #! This is the main function of the &AutoDoc; package. It can perform
-#! any combination of the following three tasks:
+#! any combination of the following tasks:
 #! <Enum>
 #! <Item>
 #!     It can (re)generate a scaffold for your package manual.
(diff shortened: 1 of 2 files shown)


gap/Magic.gd
```diff
@@ -48,6 +48,10 @@
 #!     supplementary files (such as CSS style files) into your doc directory
 #!     (see <Ref Func='CopyHTMLStyleFiles' BookName='gapdoc'/>).
 #! </Item>
+#! <Item>
+#!     It can create a maketest.g file which tests that all of the examples
```
fingolfin: should use `<F>maketest.g</F>`

gap/Magic.gd
```diff
@@ -48,6 +48,10 @@
...
+#!     in the generated documentation perform as documented.
+#! </Item>
```
fingolfin: I am really torn on whether to officially document this feature (I realize it is already mentioned in the Tutorial... *sigh*... but see issues #114, #117 and #155)

gap/Magic.gd
```diff
@@ -56,15 +60,27 @@
...
-#!     This is either the name of package, or an <C>IsDirectory</C> object.
-#!     In the former case, &AutoDoc; uses the metadata of the first package
-#!     with that name known to &GAP;. In the latter case, it checks whether
-#!     the given directory contains a <F>PackageInfo.g</F> file, and extracts
-#!     all needed metadata from that. This is for example useful if you have
-#!     multiple versions of the package around and want to make sure the
+#!     The purpose of this parameter is twofold: to determine the package
+#!     directory in which &AutoDoc; will operate, and to find the metadata
+#!     concerning the package being documented. The parameter is either a
+#!     string, an <C>IsDirectory</C> object, or omitted.
```
fingolfin: Then perhaps we should rename it from `package` to `packageOrDirectory`; or `pkgOrDir` ?

gap/Magic.gd
```diff
@@ -56,15 +60,27 @@
...
+#!     directory in which &AutoDoc; will operate, and to find the metadata
+#!     concerning the package being documented. The parameter is either a
+#!     string, an <C>IsDirectory</C> object, or omitted. 
+#!     In the first case, &AutoDoc; interprets the string as the name of a
+#!     package, uses the metadata of the first package known to &GAP;
+#!     with that name, and uses the <C>InstallationPath</C> specified in that
+#!     metadata as the package directory. The other cases directly specify the
+#!     package directory, either given by the <C>IsDirectory</C> object or as
+#!     the <C>DirectoryCurrent()</C> if the <A>package</A> parameter is
+#!     omitted. In these cases, the package directory must contain a
```
fingolfin: "package directory" -> "specified directory"

gap/Magic.gd
```diff
@@ -56,15 +60,27 @@
...
+#!     from that. The <C>IsDirectory</C> form is for example useful if you
+#!     have multiple versions of the package around and want to make sure the
 #!     documentation of the correct version is built.
 #!     <P/>
-#!     If this argument is omitted, &AutoDoc; uses the <C>DirectoryCurrent()</C>.
+#!     Note that when using <C>AutoDocWorksheet</C>
+#!     (see <Ref Sect='Chapter_AutoDoc_worksheets_Section_Worksheets' />), there
+#!     is no parameter corresponding to <A>package</A> and so the "package
+#!     directory" is always the <C>DirectoryCurrent()</C> and there is no
+#!     metadata.
```
fingolfin: @sebasguts actually, why is that? Why can't one specify a target directory for worksheets?

sebasguts: Yeah, this is somehow broken by design I would say. The `AutoDocWorksheet` command sets the package name to `AutoDocWorksheet`...

sebasguts: I would say this should probably be resolved by passing an option `Worksheet := true` to the `AutoDoc` command

gap/Magic.gd
fingolfin: Since this feature has now been deprecated, we should rather not document it.

gap/Magic.gd
```diff
@@ -51,15 +51,27 @@
...
-#!     This is either the name of package, or an <C>IsDirectory</C> object.
-#!     In the former case, &AutoDoc; uses the metadata of the first package
-#!     with that name known to &GAP;. In the latter case, it checks whether
-#!     the given directory contains a <F>PackageInfo.g</F> file, and extracts
-#!     all needed metadata from that. This is for example useful if you have
-#!     multiple versions of the package around and want to make sure the
+#!     The purpose of this parameter is twofold: to determine the package
+#!     directory in which &AutoDoc; will operate, and to find the metadata
+#!     concerning the package being documented. The parameter is either a
+#!     string, an <C>IsDirectory</C> object, or omitted.
```
fingolfin: Then perhaps we should rename it from `package` to `packageOrDirectory`; or `pkgOrDir` ?

fingolfin: Done.

Issue #60, opened 2014-08-20
Allow @Chunk on function documentation


fingolfin:
I would like to document a function, but insert its documentation at a specific point of the manual. For this, I would like to use @Chunk. But a naive test did not work. I wrote this in my GAP source:

```
#! @Chunk FOO
#! @Description
#!   Insert documentation for you function here
#! @EndChunk
DeclareGlobalFunction( "MyFunc" );
```

and then in my intro.autodoc file:

```
@InsertChunk FOO
```

But nothing appeared. But if I instead put this somewhere, its content does get inserted:

```
#! @Chunk MyChunk
#! This is some text.
#! @EndChunk
```

Issue #177, opened 2018-08-28
There is not an AutoDoc command for generating index entries


gwhitney:
While one can include GAPDoc XML in AutoDoc strings, it would still be convenient to have an AutoDoc command like `@Index Key EntryText` to generate index entries. I will post code soon that does this along with other things.

Issue #176, opened 2018-08-28
Document AutoDoc with AutoDoc and use the AutoDoc manual as a test


gwhitney:
Although the main functions are documented with AutoDoc markup, Chapters 1 and 2 of the manual are currently straight XML. Converting one or both to AutoDoc would provide a showcase of the package's functionality. Then making a test out of generating the AutoDoc manual would provide pretty good coverage of AutoDoc's functionality. Finally, the documentation could then move alongside the code it documents, helping to ensure consistency between code and documentation.

I have proposed code that, among other changes, replaces Comments.xml with Comments.autodoc, and moves the individual command documentation into gap/Parser.gi next to the implementations of the commands. It also adds generating the AutoDoc manual to the test harness started in pull request #166. I will post it soon.

Issue #175, opened 2018-08-28
(Sub)sections do not move with chunks


gwhitney:
If you create a (sub)section while a chunk is active, the (sub)section is added to its enclosing structural part at the current point in that part, not at the place where the InsertChunk occurs. This can produce very confusing results, and limits the ability to assemble a cohesive manual with a logical flow.

I recommend making (sub)sections move with the chunk they are defined in.

I have proposed code which accomplishes this, with several other changes, which I will post soon.

fingolfin:
I just run into something like this myself, it would be really useful to have this.

PR #179, opened 2018-08-28
 Recognize DeclareSynonym, with cascade of related changes

gwhitney:
This change began with recognizing DeclareSynonym and cascaded to involve a
number of related changes. First, in order for DeclareSynonym to be useful,
there must be some way to specify the manual item type for the identifier
being declared. Hence this change also implements a @ItemType command to allow
that. (Adding @ItemType also achieves the main recommendaton in issue #74.)

In documenting these two changes, it seemed best to have their documentation
in gap/Parser.gi next to their implementation, on the general principle that
documentation should be as close to the relevant code as possible. Further, in
an effort at dogfooding, it seemed best for that documentation to be in
AutoDoc format. This turned into a change in which doc/Comments.xml became
doc/Comments.autodoc and became much smaller, as most of the documentation
moved to gap/Parser.gi and gap/Markdown.gi.

Being able to use Autodoc to document most of Autodoc then entailed a few
other changes. First, Markdown-like syntax with backticks for code expressions
that would suppress the AutoDoc behavior of the quoted items was
essential. Second, it was useful to add an @index command for generating the
index entries. And third, reassembling Chapter 2 of the manual from portions
in several files necessitated allowing sections and subsections inside a chunk
to move with that chunk, and be inserted into the chapter or section at the
point at which the chunk was inserted.

(An implementation note on this last change. It was greatly facilitated by
merging the current_item ambient variable with the context_stack. The
current_item was in essence functioning as the top of the stack anyway. Having
the top of the stack in one variable and the rest of the stack in another made
some of the code a bit confusing. This change touched a large fraction of code
in the Parser, but in the end improved the locality of reference in that code
considerably, and I think made it clearer what is going on during parsing.)

In the course of these changes, it was trivial to add the ability to have code
blocks which add at the current location, with no @InsertCode necessary. So
that change is included here as well.

Finally, with so many changes, extensive testing was clearly critical. So this
commit also beefs up the worksheet.tst, and adds a new test dogfood.tst which
consists of generating the AutoDoc manual and verifying that it matches the
expected contents of the manual. I also switched to verifying the produced XML
rather than the html as the XML determines the html anyway, and this provides
a cleaner point for testing, separated from any changes in GAPDoc.

I do understand that this turned into more of a complex of changes than would
ordinarily be packed into a single commit. But this portmanteau seemed a
fairly natural stable point in which the motivating changes
(DeclareSynonym/@ItemType) are made, documented, and tested in a cohesive way.

I would be happy to do the work of separating this out into more than one
successive change to AutoDoc if that would help advance these changes. Please
just let me know what pieces you'd like it carved up into, and in what order.

Thanks very much for your consideration.

Resolves: #60, #74, #174, #175, #176, #177, #178.
It also makes further progress on #48, in adding the backtick code quotes, and
on the testing issues #57 and #152. It also touches on Issue #112, in that to
use makedoc.g in testing, I needed to add a backdoor to prevent it from
QUITting. Note also that the motivation in the original post for Issue #144
was the need to document DeclareSynonym.

This change is on top of all of my recent pull requests, except for #172, as they were also necessary in order to successfully convert Comments.xml to Comments.autodoc and have the test harness.


gap/DocumentationTree.gi
```diff
@@ -178,22 +186,26 @@ InstallMethod( DocumentationTree, [ ],
 end );
 
 ## create a chapter, section or subsection
-InstallMethod( StructurePartInTree, [ IsTreeForDocumentation, IsList ],
+## This first helper, given tree and context, returns a list of two entries.
```
fingolfin: Insert an empty line before this? Or is this intentionally joined to the preceding comment?

gap/DocumentationTree.gi
```diff
@@ -178,22 +186,26 @@ InstallMethod( DocumentationTree, [ ],
...
+## The second entry is a node to represent the structure part at that context.
+## The first is true if that structure part did not yet exist, in which case the
+## node returned as the second entry is "floating," i.e., not yet added
+## anywhere in  the tree. If the first entry is false, the node existed in the
+## tree earlier (and is presumed to have been added in some appropriate place).
+StructurePartMaybeFloating@ :=
```
fingolfin: Please don't use the horrendous GAP "namespacing" feature. Instead, add an explicit `AUTODOC_` prefix to the function name (and consider using `BindGlobal` to assign it, to avoid user's from accidentally messing with it).

gap/DocumentationTree.gi
```diff
@@ -442,12 +482,46 @@ InstallMethod( SectionInTree, [ IsTreeForDocumentation, IsString, IsString ],
     return StructurePartInTree( tree, [ chapter_name, section_name ] );
 end );
 
+## This method creates the section and if it is not already in the tree,
```
fingolfin: "... creates the section ..." what is "the" section this is referring to? Perhaps "a" section is meant?

Also, it doesn't always create a section, does it? So perhaps:  "This method returns a section object for [INSERT MEANINGFUL TEXT, like "... for the given <context>" -- but note that this example may be nonsense, I am not anymore familiar with the code here....]. If the section already exists, this is returned; otherwise, a new section node is created and returned. In the latter case, it is added as a child..."

gap/DocumentationTree.gi
```diff
@@ -442,12 +482,46 @@ InstallMethod( SectionInTree, [ IsTreeForDocumentation, IsString, IsString ],
...
+    return section_try[ 2 ];
+end );
+
 ##
 InstallMethod( SubsectionInTree, [ IsTreeForDocumentation, IsString, IsString, IsString ],
   function( tree, chapter_name, section_name, subsection_name )
     return StructurePartInTree( tree, [ chapter_name, section_name, subsection_name ] );
 end );
 
+## This method creates the subsection and if it is not already in the tree,
```
fingolfin: Same comment as above applies here.

gwhitney:
Yes, I think as you can tell from my commentary, I expected that this PR incorporated too many items to be accepted as-is/all at once. Mostly I wanted to lay out a direction that you might find worthwhile, and it was hard to elucidate each item without showing the whole collection, and plus I really did need advice on the best way to divide this into properly atomic PRs, and in what order they should be submitted. I completely agree with your proposal that the rest of the PRs should be resolved and then I should revisit this one, breaking it up along the lines you just proposed, and I will do exactly that. I just had one question about the namespacing in the comments above, which I'd be glad to hear your views on whenever convenient. Other than that, feel free to close this PR, assured that I will return to the items contained in it when the others are resolved.

fingolfin:
By now I updated all your other PRs and made them OK-to-merge (at least from my point of view). But this one is difficult. It has a ton of conflicts, and it does far too much in one commit. I don't agree with the justification for that either; it seems rather clear to me that a lot of these changes could be moved into separate commits (and even separate PRs). E.g. 
- add test suite (this code has already been merge)
- add `@ItemType` with a test
- convert `doc/Comments.xml` into `doc/Comments.autodoc`
- move parts of `doc/Comments.autodoc` into source code
- merge `current_item ambient` with the `context_stack`
- the actual fix for `DeclareSynonym`
- ability to have code blocks which are inserted at the place they appear
- ... perhaps more, I can't tell, the final commit does so many things at once ...

I tried cherry-picking just the last commit, but it seems rather hopeless at this point. So (unless you re-appear), I might try to take inspiration from this PR to re-implement a few of the above points in separate PRs.

PR #180, opened 2018-08-31
Allow AutoDoc() to take absolute dirs and run from any dir

gwhitney:
Allow AutoDoc() to succeed with a dir entry in the options record which
consists of an IsDirectory() object with an absolute path. There are two steps
to this change: First, some effort must be made to correctly compute
doc_dir_rel under such a circumstance. This commit uses the approach that if
the pkgdir is a prefix of the doc_dir, then doc_dir_rel is the remainder of
the doc_dir when the pkgdir is eliminated.

If doc_dir is absolute and pkgdir is not a prefix thereof, then there is no
sensible value for doc_dir_rel. Therefore, the second step of this change must
be to handle downstream the case that doc_dir_rel is not set. This commit
attempts to do so by presuming in such a case that there is no way to convert
the gapdoc.files to paths relative to the doc_dir, and instead converts them
to absolute paths. It may be the case that this strategy produces non-portable
documentation files, but at least MakeGAPDocDoc succeeds, and it seems to be
the only alternative in such case.

Finally, to address the need to test modifications to AutoDoc, this commit adds
a tst/testall.g that compares the results of running AutoDoc on some worksheets
with reference files for those worksheets. It also invokes a boilerplate
TestDirectory() call for maximum flexibility in adding future tests.

Those ingredients created the opportunity to test the generation of the AutoDoc
manual itself, and so this commit adds another new test, tst/dogfood.tst, which
compares the result of AutoDoc on itself with reference files.

To support this self-test, the commit also had to change makedoc.g to allow
the parameter record to be extracted by the test, and it had to allow AutoDoc
itself to be run from an arbitrary directory, not just the package directory of
the package being documented.

A minor point: this commit changes the Print() of the output directory to an
Info() call so that it can be suppressed, to ensure that test results will be
independent of the directory structure on the testing machine.

Resolves: #165.
In addition, the testing portion of this pull request seems to contribute some
progress toward #152 and/or #57.

This replaces pull request #166, which I will now close.


gwhitney:
Unfortunately, I discovered the manual test (tst/dogfood.tst) fails if the manual outdir doesn't exist prior to running the tests (I didn't discover this before because in testing, I always had the outdir around from earlier tests). So I had to push an additional 1-line change to ensure this test creates the necessary directory if need be. I know I am supposed to somehow squash the two commits now in this pull request into one, by a process which perhaps @fingolfin himself very patiently taught me several months ago, but given my infrequent use of git, I just can't seem to remember how to do this. Please accept my apologies, I am very sorry. If you could remind me how, or where to find out how, I'd be happy to do it, or I understand if you prefer to just do it yourself, or even if you want me to make a new branch  and PR "Issue165rev3" with these modifications all made in one commit (which I do by creating a new branch on upstream/master and checking it out, copying all of the files from the tip of *this* branch via `git show <THIS_BRANCH>:file > file` for all files  affected in this branch, and then doing a commit.  Somewhat klunky, but it works and is completely intuitive: start with the state you want to build on, get all of the files the way you want them to be, by whatever method, and then commit.)

Anyhow, thanks for letting me know how you'd like to proceed; I'm happy to do whatever you want/need me to.

fingolfin:
Please don't apologize for multiple commits etc. -- those are no problem for us! Even if, GitHub allows us to "squash & merge" with a single click, which merges your PR, squashed down into a single commit.

tst/dogfood.tst
```diff
@@ -0,0 +1,62 @@
+#############################################################################
+##
+##  AutoDoc package
+##
+##  Test the behavior of AutoDoc by generating its own manual
```
fingolfin: While in principle I think this is a good idea, the one drawback is that any change to the manual will lead to a test failure, unless one remember to update the dogfood test, too.

But the "fix" for that shouldn't be too hard in each instance, so I am not saying this is a major issue, just pointing it out.

tst/dogfood.tst
```diff
@@ -0,0 +1,62 @@
...
+gap> chap3good := chap3diff = Concatenation( "Files ", chap3ref, " and ", chap3, " are identical\n" );
+true
+gap> if not chap3good then Print( chap3diff ); fi;
+gap> chap4 := Filename( outdir, "_Chapter_AutoDoc.xml" );;
+gap> chap4ref := Filename( ex_dir, "_Chapter_AutoDoc.xml" );;
+gap> chap4diffout := Filename( outdir, "chap4.diff");;
+gap> command := Concatenation( "diff -s -c ", chap4ref, " ", chap4, " > ", chap4diffout );;
+gap> Exec( command );
+gap> chap4diff := ReadAll( InputTextFile( chap4diffout ) );;
+gap> chap4good := chap4diff = Concatenation( "Files ", chap4ref, " and ", chap4, " are identical\n" );
```
fingolfin: This relies on the output of `diff` being formatted in a specific way; but that may not always be valid. E.g. on a german system, commands like `diff` may have german output.

It would be much safer to check the exit code of `diff`. Unfortunately, `Exec` does not return the exit code (why????). But the same can be achieved by using `Process`. Here's a rough sketch of code (untested) doing something like that:
```gap
path := DirectoriesSystemPrograms();
diff := Filename( path, "diff" );
cmdargs:=["-s", "-c", chap4ref, chap4];
instream:=InputTextString(lines);
outstream:=TODO # could be a file stream, but could also be put into a gap string stream... or even /dev/null
exit := Process(DirectoryCurrent(), pager, instream, outstream, cmdargs);
```
This could either be put into a helper function, or used here directly.

fingolfin: Oh, and if you use a string stream for output, then you can of course also control whether and when to print it

tst/testall.g
```diff
@@ -0,0 +1,105 @@
...
+            all_ok := false;
+            continue;
+        fi;
+        dif := Concatenation( af, ".diff" );
+        command := Concatenation( "diff -s -c ", exf, " ", af, " > ", dif );
+        Exec( command );
+        report := ReadAll( InputTextFile( dif ) );
+        desired := Concatenation( "Files ", exf, " and ", af,
+                                  " are identical\n" );
+        if report <> desired then
```
fingolfin: The remarks on `Diff` from above apply here, too. Perhaps put it into an `AUTODOC_Diff` helper function, to be used in both places.

Then, I'd consider moving the code in the *body* of this loop into a helper function, too (which takes `ws` as argument). Afterwards, this file gets much shorter.
Finally, the loop could be moved *out* of testall.g, into a `.tst` file of its own. Then we are back to a "pure" testall.g".

fingolfin:
BTW, I am also experimenting with enabling Travis & Codecov support in PR #182.

tst/dogfood.tst
fingolfin: BTW, instead of `-c`, personally I'd use `-u`.

fingolfin: I have written some code for this now, and will shortly make a PR. Stay tuned.

fingolfin:
Based on this PR, I created https://github.com/gap-packages/AutoDoc/pull/183.

fingolfin:
This is now obsolete, with #183 merged.

PR #181, opened 2018-08-31
Make @Begin.../@EndExampleSession respect plain_text_mode

### LINKED ISSUE #167: @BeginExampleSession/@EndExampleSession do not respect plain_text_mode

gwhitney:
In order to uniformize the handling of #! prefixes in AutoDoc files,
gap/Parser.gi is changed to pass lines between @BeginExampleSession and
@EndExampleSession through unchanged when plain_text_mode is true.

Note this commit also supplies two test worksheets. They will not have any
effect unless/until the test harness supplied with pull request #180 is put in
place. But the worksheets do at least serve as illustrations of the intended
behavior post this change, and should otherwise be innocuous.

Resolves #167 

This replaces pull request #168, which I will now close.

### DIFF
--- a/gap/Parser.gi
+++ b/gap/Parser.gi
@@ -474,14 +474,12 @@ InstallGlobalFunction( AutoDoc_Parser_ReadFiles,
         od;
         return example_node;
     end;
-    read_session_example := function( is_tested_example )
-        local temp_string_list, temp_curr_line, temp_pos_comment, is_following_line, item_temp, example_node;
+    read_session_example := function( is_tested_example, plain_text_mode )
+        local temp_string_list, temp_curr_line, temp_pos_comment,
+              is_following_line, item_temp, example_node,
+              incorporate_this_line;
         example_node := DocumentationExample( tree );
-        if is_tested_example = false then
-            example_node!.is_tested_example := false;
-        else
-            example_node!.is_tested_example := true;
-        fi;
+        example_node!.is_tested_example := is_tested_example;
         temp_string_list := example_node!.content;
         while true do
             temp_curr_line := ReadLineWithLineCount( filestream );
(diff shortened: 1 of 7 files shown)


gap/Parser.gi
```diff
@@ -438,13 +438,19 @@ InstallGlobalFunction( AutoDoc_Parser_ReadFiles,
...
+                temp_pos_comment := PositionSublist( temp_curr_line, "#!" );
+                if temp_pos_comment <> fail then
+                    incorporate_this_line := true;
+                    temp_curr_line := temp_curr_line{[ temp_pos_comment + 2 .. Length( temp_curr_line ) ]};
+                    if Length( temp_curr_line ) >= 1 and temp_curr_line[ 1 ] = ' ' then
+                        Remove( temp_curr_line, 1 );
+                    fi;
                 fi;
+            fi;
+            if incorporate_this_line then
```
fingolfin: I am not very familiar with this code, so @sebasguts should review this part of the PR.

gap/Parser.gi
```diff
@@ -420,8 +420,8 @@ InstallGlobalFunction( AutoDoc_Parser_ReadFiles,
         od;
         return example_node;
     end;
-    read_session_example := function( is_tested_example )
-        local temp_string_list, temp_curr_line, temp_pos_comment, is_following_line, item_temp, example_node;
+    read_session_example := function( is_tested_example, pt_mode )
```
fingolfin: pt_mode = plaintext_mode?

If so, please rename it, or at least add a comment here that states what it is. This function is already cryptic enough, no need to make it worse ;-)

fingolfin:
@gwhitney any plans on your side to update this PR, so that we can get it merged (right now, there is only a minor comment by me, but this PR at the very least should be rebased and may need minor adjustments for the tests; and hopefully, @sebasguts would then also review it.

gap/Parser.gi
fingolfin: @sebasguts could you review this PR? I have now updated it (mostly the tests)

PR #188, opened 2019-02-19
Added possibility to mark function arguments via {A}

sebasguts:
And, on the way, make markdown support different markers
for beginning and end of a tag


fingolfin:
The broken Travis test is my fault; it should work once Travis cycles again.

Issue #189, opened 2019-02-20
commit 6f7dfccda4cc38bd325cb1b59cb5912947c6568d breaks many packages


zickgraf:
After commit 6f7dfccda4cc38bd325cb1b59cb5912947c6568d, `make -B doc` in `CAP_project/CAP` leads to an error:

```
XML Parse Error: Line 536 Character 93
Original file: /home/xxxxxx/.gap/pkg/CAP_project/CAP/doc/_Chapter_Category_of_Categories.xml, line number 141.
-----------
where <Math>D_i = A_i</Math> if <Math>b_i = \mathtt<A>false</A></Math>, and <Math>D_i = A_i^<A>\mathrm<A>op</A></A></Math> otherwise.
                                                                                            ^
-----------
Wrong element in A: A
!!! Type 'Show();' to watch the input string in pager - starting with
    line containing error !!!
Or 'ShowOrigin();' to look it up in its source file.
Error,  at /home/xxxxxx/.gap/pkg/GAPDoc/lib/XMLParser.gi:146 called from
ParseError( str, r.start, Concatenation( "Wrong element in ", name, ": ", r.content[Position( l, false )].name ) ); at /home/xxxxxx/.gap/pkg/GAPDoc/lib/GAPDoc.gi:99 called from
CheckAndCleanGapDocTree( x, str ) at /home/xxxxxx/.gap/pkg/GAPDoc/lib/GAPDoc.gi:157 called from
func( elm ) at /usr/lib/gap/lib/coll.gi:1552 called from
ForAll( r.content, function ( x )
      return CheckAndCleanGapDocTree( x, str );
  end ) at /home/xxxxxx/.gap/pkg/GAPDoc/lib/GAPDoc.gi:157 called from
CheckAndCleanGapDocTree( x, str ) at /home/xxxxxx/.gap/pkg/GAPDoc/lib/GAPDoc.gi:157 called from
...  at makedoc.g:20
you can 'quit;' to quit to outer loop, or
you can 'return;' to continue
```

This also happens in the following packages:
* https://github.com/homalg-project/FinGSetsForCAP/
* https://github.com/homalg-project/Bialgebroids/
* https://github.com/homalg-project/FunctorCategories/

Is this an intended backwards-incompatible change?

sebasguts:
@zickgraf Thank you for reporting!
@fingolfin I chose `{}` for arguments, which was super duper bad apparently.
[...]

PR #190, opened 2019-02-20
Changed {} to § in Markdown for function arguments

sebasguts:


oysteins:
Is it necessary to have `§` at both the beginning and the end of an argument, not just the beginning?  An argument name must be a valid GAP identifier, so it should be possible to detect where it ends.

sebasguts:
Indeed, that would be possible, but a bit harder to implement. I will see if I get it done quickly, but first I want to have a new release that fixes #189

sebasguts:
@fingolfin allright, though

sebasguts:
@mtorpey suggested to use double-backticks `` for arguments.
@oysteins @fingolfin What do you think?
Another option would be to do `$foo`, without a closing `$`

mtorpey:
Hi folks!

As @sebasguts says, my suggestion would be double backticks for arguments (`<A>` tags) and single backticks for other code (`<C>` tags), as shown:
```
If the argument ``inter`` is set to `true`, then an interactive prompt will be created.
```
I believe this would keep the code clean, readable, and would not introduce any unnecessarily heavy distinction between normal code and arguments.  The point of Markdown, for me, is that it just looks like readable plaintext and not a markup language at all, and symbols like `@` and `§` and `$` tend to spoil that a little bit.

That said, I'd be happy with a symbol like `$` either at the beginning, or at the beginning and end -- this would still be nice-looking enough, and certainly much better than `<A></A>`, which is the best option at the moment.

I'm not a fan of `§`, partly because it's non-ASCII, but also because it's difficult to type on silly British keyboards.

alex-konovalov:
Backticks for `<C>` elements would be great, so I very much agree with @mtorpey that we should keep them for this purpose. In this case, double backticks for `<A>` sounds good for me too. 

I don't like `$` since many of us will unconsciously parse it to LaTeX notation, so that will ultimately cause confusion.

fingolfin:
This PR obviously needs to be rebased now.

I am a bit concerned about the idea of using double backticks, because I fear this might introduce a serious incompatibility with regular MarkDown, and I am still dreaming of us using a proper MarkDown parser with minor extensions at some point (e.g. by adapting <https://github.com/commonmark/cmark>).

I share the concern about $ -- intuitively I'd expect that to be an alias for `<M> ... <M>`.

And then § is indeed not ASCII, which is also problematic, though the more important concern here is how difficult it is to actually type it.

One thing we could do is that if we see a single backtick quoted identifier, we could lookup whether it is an argument name, and then use that to decide whether to emit `<C>` or `<A>`. I'd expect that cover 95% of all cases just fine...?

Finally, and in the end, I just think that typing `<A>foo</A>` is not *that* terrible... ;-)

mtorpey:
I like the sound of single backticks for both <C> and <A>, with the computer working it out from context.  The fewer technical things we have to worry about when writing the doc, the better!

sebasguts:
Yep, it needs to be rebased, but atm I am not sure if we should rather close it for now, or come up with something sensible. To be honest, in most cases we know the identifiers for arguments. So maybe we should just look for them and put them in `<A>` tags, without any "markdown" indication? This can of course cause more troubles.

alex-konovalov:
@sebasguts This will cause massive problems when identifiers are real words.

sebasguts:
Indeed, that's why I am quite reluctant to do it. We could also use backticks for identifiers, and if they are identifiers, we make them `<A>` instead of `<C>`. What about this?

alex-konovalov:
@sebasguts by "identifiers" you mean "arguments"? You can also check if something is a keyword, and convert ``false`` to `<K>false</K>` etc.

sebasguts:
@alex-konovalov Yep, by identifiers I mean argument names. I will also do the keywords. I guess we can trust `GAPInfo.Keywords` here?

fingolfin:
So I guess this can essentially be closed? Perhaps after extracting an issue or two to remind us about the content and result of the discussion?

fingolfin:
Just filed issue #207, so now closing this PR

PR #191, opened 2019-02-23
Add extract_examples option

### LINKED ISSUE #114: Alternative to maketest?

### LINKED ISSUE #117: Replace `maketest` option by a new AutoDocRunExamples() function

### LINKED ISSUE #155: Using maketest.g in package standard tests

fingolfin:
If this option is passed to AutoDoc(), then all examples in the manual
are extracted into .tst files, which are placed into the tst/ directory
of the package.

This way, all manual tests can be run via the same `TestDirectory` invocation
used for all other tests. And the test can be regenerated from the manual
as part of a package's `makedoc.g` script, which in turn means they are
regenerated when a package release is made using ReleaseTools.

WIP: before this is merged, documentation needs to be added. But I wanted to solicit feedback first, in case it leads to major changes which then in turn might affect the to-be-written documentation.

One thing to argue about is where to place the `.tst` files: for the old TeX based gapmacros system, they were placed into `doc`. One potential advantage of doing that: one could run these tests with different testoptions from those employed for all other tests. E.g. for manual tests, one might want to ignore whitespace.

But for my personal use, this is not really necessary... Not sure what others think, though?

Resolves #114
Resolves #117
Resolves #155


sebasguts:
This looks good, thank you very much.

I do not think it is necessary to put those tests into `/doc`, `/tst` is fine. We can always prefix the `tst` files, like `doctest_...`, that way, one could distinguish them in the `testall.g` if one really wants to.

There currently is the `maketest` option record. Shouldn't we place that option inside there?

fingolfin:
Frankly, I never really understood or used `maketest`, and did not want to have to figure it out; it seems to be for a feature which is kind of orthogonal to what I am adding here, which is why I chose a separate option name for it. That also simplified the code, as I did not have to modify the existing `maketest` code paths.

fingolfin:
In fact, for most purposes, it seems to me that the code added here is superior to `maketest`, and I'd wonder if anybody using `maketest` might not want to switch to `extract_examples` at some point (right now, it seems that only these packages use maketest:
* Homalg and its subpackages
* CAP and its subpackages
* images
* ArangoDBInterface
* NormalizInterface
* (SingularInterface)

sebasguts:
@fingolfin It is indeed superior, and all packages should switch.

So then lets have the different option record, as we might want to have options at some place (folder specs, etc).

sebasguts:
We could even go ahead and make the current maketest option just use this option, and then make `maketest.g` call the right `.tst` files.

alex-konovalov:
I have adapted one of homalg packages in https://github.com/homalg-project/homalg_project/pull/233 to use their existing `maketest.g` as default test of the package. But if this will be accepted, I am certainly in favour of this approach. I think having physical test files placed into `tst` is better than generating them on fly.

fingolfin:
This PR is now ready for proper review.

gap/AutoDocMainFunction.gi
```diff
@@ -419,3 +419,65 @@ InstallGlobalFunction( CreateMakeTest,
...
+        output := OutputTextFile( name, false ); # to empty the file first
+        SetPrintFormattingStatus( output, false ); # to avoid line breaks
+        ch := tst[i];
+        AppendTo(output, "# ", pkgname, ", chapter ",i,"\n");
+        AppendTo(output, "#\n",
+                         "# DO NOT EDIT THIS FILE - EDIT EXAMPLES IN THE SOURCE INSTEAD!\n",
+                         "#\n",
+                         "# This file has been autogenerated with GAP. It contains examples\n",
+                         "# extracted from the documentation. Each example is preceded by the\n",
+                         "# comment which points to the location of its source.\n",
```
sebasguts: I am not quite sure about adding the source here.

In most AutoDoc cases, the source will be a generated `.xml` file, so I am not sure how useful it would be. It might even irritate people not aware of the fact that AutoDoc generates XML-files.

fingolfin: Having the origin of each test is super useful for people who just use AutoDoc to convert GAPDoc XML inputs. Of course it is less useful if one uses full-blown AutoDoc, including for the inputs; but I'd argue even then it can be helpful, as by looking at the XML, I can usually at least identify some context 

Anyway, I wouldn't mind if we only inserted the "source location" for non-AutoDoc-XML files (to keep things simple, that would mean excluding sources starting with an `_`). But is it really worth the effort to strip this information? I don't quite understand what is irritating about it ... at least if we were to adjust the comment to clarify that if the XML "source" is generated, then of course it is not the *actual* source...

sebasguts: Adjusting the comment would be sufficient, yes. I just assume it might be irritating for people who have not really seen the xml files before.

fingolfin: @sebasguts I've update this text now. Better?

fingolfin:
BTW, an immediate "obvious" improvement to this function is to allow a record instead of `true`, and then in that record allow specifying the output directory, if one does not want to output to `tst`.

gap/Magic.gd
```diff
@@ -323,11 +323,17 @@
...
 #!         </List>
 #!     </Item>
-## This is the maketest part. Still under construction.
+#!     <Mark><A>example_test</A></Mark>
+#!     <Item>
+#!         If set to <K>true</K>, then all manual examples are extracted and placed
+#!         into files <F>tst/PACKAGENAME01.tst</F>,  <F>tst/PACKAGENAME02.tst</F>, ...
+#!         and so on, with one file for each chapter.
+#!     </Item>
+## TODO: should we mark maketest as deprecated and recommend extract_examples instead?
```
sebasguts: I think we should, yes.
Add the beginning, we could write something like:
> This option is deprecated. To be more compatible with the GAP test framework, please use `example_test`.

fingolfin: I've added some text in this vein. Feel free to merge.

sebasguts:
@fingolfin The tests show a diff. Do you want to address it, or should I try to fix it?

fingolfin:
I am busy with child care for the next couple hours, so if you could take care of it...

fingolfin:
@sebasguts tests fixed now

sebasguts:
Sorry, was already afk yesterday.

PR #192, opened 2019-03-01
Add AUTODOC_FormatDate helper

fingolfin:
This way, other code can format dates in a "human friendly" fashion, e.g. for use in XML entities for the package date.

Perhaps something like this should instead be added to GAP itself, or the utils package, but we can start here for now.

### DIFF
--- a/tst/worksheets/general.expected/title.xml
+++ b/tst/worksheets/general.expected/title.xml
@@ -6,6 +6,6 @@
     General Test
   </Title>
   <Date>
-    2018 August 30
+    30 August 2018
   </Date>
   </TitlePage>
--- a/tst/worksheets/general.sheet/worksheet.g
+++ b/tst/worksheets/general.sheet/worksheet.g
@@ -11,7 +11,7 @@
 Print( "Pretend this is a code file.\n" );
 Print( "(Even though we never use it that way.\n" );
 #! @Title General Test
-#! @Date 2018/08/30
+#! @Date 30/08/2018
 #! @Chapter SomeChapter
 #! This is dummy text
 #! @BeginExampleSession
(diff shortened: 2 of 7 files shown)


tst/worksheets/general.sheet/worksheet.g
```diff
@@ -11,7 +11,7 @@
 Print( "Pretend this is a code file.\n" );
 Print( "(Even though we never use it that way.\n" );
 #! @Title General Test
-#! @Date 2018/08/30
+#! @Date 30/08/2018
```
fingolfin: The new checks actually caught a bug in our test suite :)

fingolfin:
Oops, tests passed when they should not have...

fingolfin:
I added a commit which ensures that in the future, `AUTODOC_TestWorkSheet` will fail if any diffs are found, verified that this now indeed is the case, and then fixed the broken diff.

I *think* this PR is now good to go.

tst/worksheets/general.expected/title.xml
```diff
@@ -6,6 +6,6 @@
     General Test
   </Title>
   <Date>
-    2018 August 30
+    30 August 2018
```
sebasguts: I honestly do not like the way this is displayed now (but also not the way it was displayed before). I think the most natural way I can think of for me seems `August 30, 2018`. But this is probably an opinion and bikeshedding, so we can leave it this way. It is probably better than the old way already.

fingolfin: Actually, that's the difference between wrong (as it was), british (the way it is now) and american (the way you prefer). See also e.g. https://www.englishclub.com/vocabulary/time-date.htm

sebasguts: Okay, that explains it quite well. I am just use to see the American formatting all the time. But as said, I do not care too much here, I just wanted to point it out. Lets merge this :)

PR #197, opened 2019-03-13
Add option extract_examples.units and pass it to ExtractExamples

zickgraf:
This option allows to create one .tst file per section, subsection or example instead of per chapter.

I have tried to not change any existing behaviour. That's why I always pad the integer inserted into the filename to at least 2 digits.


sebasguts:
@zickgraf Please have a look at the travis tests, it seems like you did not update the expected manual xml files for the tests, so the travis tests fail.

Otherwise, good work, thank you very much.

The only thing I am not sure about are the file names. Producing somehow more useful filenames is probably too hard, and not necessary. However, it does not seem like GAPDoc would output any useful information (Section/Chapter name or anything) with the example tree, so we can just leave it this way for now.

zickgraf:
Ah, sorry, I didn't realize AutoDoc verifies its own documentation and since I did not change the default behaviour I expected the tests to pass. I hope this is fixed now.

I have also thought about filenames. The only idea I had was the following: we know the source xml file and the line where the example starts. So we could just look for the last chapter, section, subsection or heading preceding the example and use this as the filename. Two problems come to my mind immediately:
1. There might be sections in different chapters with the same name (or even in the same chapter, but then I would blame the user). One could either simply append a number or one could prefix them with the name of the chapter. However, the latter case would lead to really long filenames I guess, especially when expanding this to subsections (and for "Single" this would still not be enough).
2. We would have to generate a list with forbidden characters for filenames across different filesystems. Is there existing code in GAP which might cover this already?

If you like the idea in principle, I can look into this but would like to split this off in a separate PR.

fingolfin:
How about just using the chapter and/or section numbers?  Something like this:  `ch12_sec7.tst`;  or perhaps `ch012_sec007.tst` for better sorting. Can also be longer. This way, it is still reasonably easy to identify where each example comes from, but the filenames stay short.

zickgraf:
I don't see how I could access the number of the chapter, section or subsection the example is in.

fingolfin:
Hrm, I see. Once again it is really annoying that we only act as a frontend for GAPDoc, instead of having control over the whole stack... Perhaps we need to think hard about that at one point in the future...

sebasguts:
Looking at GAPDoc, we could also recreate the functionality of the [ExtractExamplesXMLTree](https://github.com/frankluebeck/GAPDoc/blob/master/lib/Examples.gi#L105) function, which seems reasonable. The function would have not only chapter and section numbers, but also the titles at their hands (as the xmltree contains all this information). Should we write our own extract examples, which needs to be kept up-to-date with the changes in GAPDoc?

gap/Magic.gd
```diff
@@ -323,15 +323,30 @@
 #!
 #!         </List>
 #!     </Item>
-#!     <Mark><A>example_test</A></Mark>
+#!     <Mark><A>extract_examples</A></Mark>
```
fingolfin: Ouch, so the manual was actually wrong here... Thanks for catching and fixing this.

PR #198, opened 2019-03-14
Provide `@NotLatex` in addition to `@LatexOnly`

### LINKED ISSUE #186: Provide `@NotLatex` in addition to `@LatexOnly`

zickgraf:
Fixes #186

Note: I have no deep understanding of the parser code. I have just copied the `LatexOnly` parts and in my tests everything works.

### DIFF
--- a/doc/Comments.xml
+++ b/doc/Comments.xml
@@ -474,7 +474,7 @@ i := i + 1;
 </Subsection>
 
 <Subsection Label="@LatexOnly">
-<Index Key="@LatexOnly"><C>@@LatexOnly, @BeginLatexOnly, and @EndLatexOnly</C></Index>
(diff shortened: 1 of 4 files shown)


fingolfin:
Coverage drops because of this. It would be nice if there was a test for this, e.g. by editing `tst/worksheets/general.sheet/worksheet.g` suitably, then updating the content of `tst/worksheets/general.expected/`  accordingly (we probably should provide a script to take care of the latter... I'll file an issue)

fingolfin:
Anyway, I am also fine with merging this as-is, and then we add tests later. We don't test many other things right now either, so who am I to complain :-/

zickgraf:
I have added tests for both `LatexOnly` and `NotLatex` and hope that this will increase the coverage.

sebasguts:
Very nice!

PR #201, opened 2019-03-28
Only add package authors to manual title page; add some contributors to PackageInfo.g

### LINKED ISSUE #38: Distinguish between authors and maintainers?
Currently, when generating the title page, we do not distinguish between authors and maintainers (note that these two properties are actually independent). Perhaps we should add some visual cue to distinguish them? Or have two lists, one for authors, one for maintainers (resp. for maintainers that are not also authors) ? Or add a little symbol or text for each <Author> indicating whether it is an author, maintainer or both?

Or perhaps this is unimportant and we should just ignore this.

fingolfin:
Resolves #38

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -2,6 +2,7 @@ This file describes changes in the AutoDoc package.
 
 2019.MM.DD
   - Add opt.extract_examples to AutoDoc function
+  - Add @NotLatex command to complement @LatexOnly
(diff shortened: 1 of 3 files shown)


fingolfin:
Rebased to resolve conflicts -- @sebasguts please re-approve and merge :-)

fingolfin:
Actually, I'll also add Fabian to the contributors list.

Issue #157, opened 2018-03-26
A few suggestions for the AutoDoc manual


ThomasBreuer:
- [x] 2.1.8 `@ChapterInfo`:
    GAPDoc's `<A>` element is intended only for arguments.
    In the given example, `p` and `G` and `ordtbl` do not occur as
    arguments, they could be marked with `<M>` or `<C>` elements.

- [x] 2.2.3 `@EndSection`:
    The text suggests that putting text "between sections" of a chapter
    makes sense.
    I would recommend that a chapter might have some introductory text
    before the first section starts (so this text belongs to the chapter
    but not to a section), but nothing should appear between sections.

- [x] 2.2.5 `@EndSubsection`:
    See the above comment on `@EndSection`.

- [x] 2.2.7 `@EndAutoDoc`:
    Write "effect" not "affect"?

- [x] 2.2.10 `@Level`:
    Write "about levels" not "about groups"?

- [x] 2.2.13 `@BeginExampleSession and @EndExampleSession`:
    Write "GAP prompt" instead of "gap prompt"?
    (Alternatively, one could write "`gap> `" prompt,
    but there is also the partial prompt "`> `".)

- [ ] 2.2.14 `@BeginLog and @EndLog`:
    References to the GAPDoc manual could be made more specific,
    and could appear as links.

- [x] 2.3 `Title page commands`:
    Write "from <F>PackageInfo.g</F>" and
    "do not have a <F>PackageInfo.g</F> file to extract".

fingolfin:
Unfortunately we (well, at least I) had completely forgotten about this issue. I now addressed some of the points listed here in PR #204 (and marked them in the issue description).

fingolfin:
I also mostly agree with @oysteins analysis and suggestions. However, I'd keep `@EndSection` and `@ @EndSubsection` around for backwards compatibility, I'd just change them to (a) print a fat warning, and (b) probably they should do nothing.

This fits in very well with the Markdown spirit, too, where we might use `# Chapter`, `## Section` and `### Subsection` to start the various things, but there is no way to "end" them other than to start a new chapter/section/subsections/

fingolfin:
@sebasguts any additional thoughts or comments?

fingolfin:
I just filed issue #206 which should take care of the EndSection and EndSubsection remarks. This leaves the remark on @BeginLog and @EndLog where we should indeed consider adding explicit links to the GAPDoc manual (actually, this would be nice in other places, too)

PR #204, opened 2019-03-28
Improve manual

fingolfin:
- address parts of issue #157
- use &AutoDoc; entity in some additional spots
- replace @ExampleSession by @BeginExampleSession in the documentation
  for @BeginExampleSession
- use <C>...</C> when referring to AutoDoc @-commands
- improve index entries for @-commands
- a bunch of minor typos and tweaks of the text

### DIFF
--- a/doc/Comments.xml
+++ b/doc/Comments.xml
@@ -30,24 +30,24 @@ this, there are several commands you can put into the &AutoDoc; comment
 before the declaration. Currently, the following commands are provided:
 
 <Subsection Label="@Description">
-<Index Key="@Description"><C>@Description</C></Index>
-<Heading>@Description <A>descr</A></Heading>
+<Index Key="@Description"><C>@Description <A>descr</A></C></Index>
+<Heading><C>@Description <A>descr</A></C></Heading>
 Adds the text in the following lines of the &AutoDoc; to the description
 of the declaration in the manual. Lines are until the next &AutoDoc; command
 or until the declaration is reached.
 </Subsection>
 
 <Subsection Label="@Returns">
-<Index Key="@Returns"><C>@Returns</C></Index>
-<Heading>@Returns <A>ret_val</A></Heading>
+<Index Key="@Returns"><C>@Returns <A>ret_val</A></C></Index>
+<Heading><C>@Returns <A>ret_val</A></C></Heading>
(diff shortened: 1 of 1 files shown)


doc/Comments.xml
```diff
@@ -116,9 +116,9 @@ As an example, a full &AutoDoc; comment with all options could look like this:
 <Listing><![CDATA[
 #! @Description
 #! Computes the list of lists of degrees of ordinary characters
-#! associated to the <A>p</A>-blocks of the group <A>G</A>
-#! with <A>p</A>-modular character table <A>modtbl</A>
-#! and underlying ordinary character table <A>ordtbl</A>.
+#! associated to the <M>p</M>-blocks of the group <M>G</M>
```
sebasguts:
```suggestion
#! associated to the $p$-blocks of the group $G$
```
I am not sure if this is reasonable or should not be shown here.

doc/Comments.xml
```diff
@@ -116,9 +116,9 @@ As an example, a full &AutoDoc; comment with all options could look like this:
...
+#! with <M>p</M>-modular character table <A>modtbl</A>
```
sebasguts:
```suggestion
#! with $p$-modular character table <A>modtbl</A>
```

doc/Comments.xml
```diff
@@ -116,9 +116,9 @@ As an example, a full &AutoDoc; comment with all options could look like this:
...
+#! and underlying ordinary character table <C>ordtbl</C>.
```
sebasguts:
```suggestion
#! and underlying ordinary character table `ordtbl`.
```

doc/Comments.xml
fingolfin: Not sure what you mean with that... ? Are you wondering whether to use `$` here or not? Well, my personal stance is that I like your suggestion, after all, we want to show people how "true" AutoDoc comments look like. So, once we implement single backtick support with automagic distinction between arguments, keywords, code, I'd also not hesitate to use those here instead of `<A>` and `<C>`

sebasguts: Yep, you are right. I was just not sure if we should use markdown in one of the first examples, but you are right, we should directly show people how to use AutoDoc

PR #208, opened 2019-04-13
Replace scaffold.gapdoc_latex_options by gapdoc.LaTeXOptions

### LINKED ISSUE #149: Move gapdoc_latex_options to gapdoc record
and rename it to `GAPDocLaTeXOptions`. Then mark `gapdoc_latex_options` as deprecated and provide an example for `GAPDocLaTeXOptions`.

cdwensley:
This is a first attempt to address issue #149. 
After these changes the following makedoc.g files appear to run correctly: (1) after changing to GAPDocLatexOptions - AutoDoc; XMod,  (2) still using gapdoc_latex_options: groupoids; intpic; factint; sgpviz, xmodalg.  The makedoc.g in numericalsgps throws a lot of 'inputenc Errors', but that may not be due to this PR? 
Question: should the entities also be moved from scaffold to gapdoc? 

Resolves #149

### DIFF
--- a/gap/Magic.gd
+++ b/gap/Magic.gd
@@ -178,13 +178,6 @@
 #!             Replaces the standard header from &GAPDoc; completely with the header in this LaTeX file.
 #!             Please be careful here, and look at GAPDoc's latexheader.tex file for an example.
 #!         </Item>
-#!         <Mark><A>gapdoc_latex_options</A></Mark>
-#!         <Item>
-#!             Must be a record with entries which can be understood by SetGapDocLaTeXOptions. Each entry can be a string, which
-#!             will be given to &GAPDoc; directly, or a list containing of two entries: The first one must be the string "file",
-#!             the second one a filename. This file will be read and then its content is passed to &GAPDoc; as option with the name
-#!             of the entry.
-#!         </Item>
 #!
 #!         </List>
 #!     </Item>
@@ -296,6 +289,17 @@
 #!             <Br/>
 #!             <E>Default value: <C>[ ".", "gap", "lib", "examples", "examples/doc" ]</C>.</E>
 #!         </Item>
+#!
+#!         <Mark><A>LaTeXOptions</A></Mark>
+#!         <Item>
+#!             Must be a record with entries which can be understood by
+#!             <Ref Func='SetGapDocLaTeXOptions' BookName='gapdoc'/>. Each entry can be a
+#!             string, which will be given to &GAPDoc; directly, or a list containing of
+#!             two entries: The first one must be the string "file", the second one a
+#!             filename. This file will be read and then its content is passed to &GAPDoc;
+#!             as option with the name of the entry.
+#!         </Item>
+#!
 #!         <Mark><A>gap_root_relative_path</A></Mark>
 #!         <Item>
 #!             Either a boolean, or a string containing a relative path.
--- a/gap/Magic.gi
+++ b/gap/Magic.gi
@@ -72,7 +72,7 @@ function( arg )
           doc_dir, doc_dir_rel, tmp, key, val, file,
           pkgdirstr, docdirstr,
           title_page, tree, is_worksheet,
-          position_document_class, gapdoc_latex_option_record,
+          position_document_class,
           makeDocFun, args;
 
     if Length( arg ) >= 3 then
@@ -408,8 +408,6 @@ function( arg )
     #
     # Generate scaffold
     #
-    gapdoc_latex_option_record := rec( );
-
     if IsBound( scaffold ) then
         ## Syntax is [ "class", [ "options" ] ]
         if IsBound( scaffold.document_class ) then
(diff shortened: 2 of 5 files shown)


cdwensley:
dogfood.tst fails - and I'm not sure what it is supposed to be doing. 
Question: why 'GAPDocLaTeXOptions' when GAPDoc has 'SetGapDocLaTeXOptions'?

fingolfin:
hi @cdwensley, thank you for this PR! And my apologies for reacting only now, I only saw it today by chance (well, and I've been busy, with a new since 1st of April etc.).

Thank you for pointing out that GAPDoc uses `SetGapDocLaTeXOptions` -- I think that we therefore should use `GapDocLaTeXOptions` and not `GAPDocLaTeXOptions`.

I'll comment more once I had a chance to look at the actual content of your PR.

gap/Magic.gd
```diff
@@ -2,7 +2,7 @@
 ##
 ##  AutoDoc package
 ##
-##  Copyright 2012-2016
+##  Copyright 2012-2019
```
fingolfin: Please remove this from this PR, it doesn't belong in there (and it's conflicting with PR #210, which makes this change superfluous).

gap/Magic.gi
```diff
@@ -293,6 +293,13 @@ function( arg )
         gapdoc := rec();
     fi;
 
+    #
+    # Allow for legacy gapdoc_latex_options
+    #
+    if IsBound( scaffold.gapdoc_latex_options ) and not IsBound( gapdoc.GAPDocLaTeXOptions ) then 
+        gapdoc.GAPDocLaTeXOptions := scaffold.gapdoc_latex_options;
```
fingolfin: I would suggest to also issue a warning here, say `Info(InfoWarning, 1, "Please replace the DEPRECATED option <scaffold.gapdoc_latex_options> by <gapdoc.GapDocLaTeXOptions>");`

fingolfin: Perhaps we also call it just `gapdoc.LaTeXOptions` instead? The extra `GapDoc` seems redundant? Perhaps @sebasguts also has an opinion on this?

gap/Magic.gi
```diff
@@ -2,7 +2,7 @@
 ##
 ##  AutoDoc package
 ##
-##  Copyright 2012-2016
+##  Copyright 2012-2019
```
fingolfin: Also needs to be removed

gap/Magic.gd
```diff
@@ -183,7 +183,7 @@
 #!             Replaces the standard header from &GAPDoc; completely with the header in this LaTeX file.
 #!             Please be careful here, and look at GAPDoc's latexheader.tex file for an example.
 #!         </Item>
-#!         <Mark><A>gapdoc_latex_options</A></Mark>
+#!         <Mark><A>GAPDocLaTeXOptions</A></Mark>
```
fingolfin: This needs to be moved from the `scaffold` to the `gapdoc` section, at least if it is supposed to fit with the rest of this PR.

cdwensley:
All requested changes added, including replacing gapdoc.GAPDocLaTeXOptions by gapdoc.LaTeXOptions and moving the latex_header_file item to the gapdoc section - but exactly where in this section should it go?  
This is all getting beyond my GitHub competency - I'd be very happy if other people did any editing that was still necessary.

fingolfin:
I've updated this PR (rebased on latest master, fixed some issues with it, made the tests pass, update the commit message).

gap/Magic.gd
```diff
@@ -296,6 +289,17 @@
 #!             <Br/>
 #!             <E>Default value: <C>[ ".", "gap", "lib", "examples", "examples/doc" ]</C>.</E>
 #!         </Item>
+#!
+#!         <Mark><A>LaTeXOptions</A></Mark>
```
fingolfin: We are kind of split in whether we use `snake_case` or `CamelCase` for option record entries... Well, the former usually were added by @sebasguts and the latter by me :-). It is a bit jarring that we now have `gapdoc.LaTeXOptions` right next to `gapdoc.gap_root_relative_path` ... :-/. But of course this is not new, i.e., not a principle issue with this PR...

IMHO the major argument for using `CamelCase` is that this is what GAP uses, period. Sure, `snake_case` is popular in Python and Julia and other languages -- but not in GAP.

That said, I can also live with use `snake_case`. It would simply be nice if we made these things uniform. Eventually. Not in this PR, just in general :-)

So perhaps we should decide on one of the two conventions, and then convert everything for that (of course while retaining support for the old versions for the foreseeable future, possibly forever).

gap/Magic.gd
```diff
@@ -296,6 +289,17 @@
...
+#!         <Item>
+#!             Must be a record with entries which can be understood by
+#!             <Ref Func='SetGapDocLaTeXOptions' BookName='gapdoc'/>. Each entry can be a
+#!             string, which will be given to &GAPDoc; directly, or a list containing of
+#!             two entries: The first one must be the string "file", the second one a
+#!             filename. This file will be read and then its content is passed to &GAPDoc;
+#!             as option with the name of the entry.
```
fingolfin: Wait, what?!? I never knew we supported that... else I'd have loudly complained to @sebasguts ... :-). This seems to be a pointless complication, as one can simply use `StringFile` to get that same effect, no? I just had a quick look and found no package using this, so I hope we can remove this (in another PR! This is not a change request for the PR under review!)

gap/Magic.gd
sebasguts: We could actually just support arbitrary versions by converting everything to lower case without underscores.

However, I think GAPDoc likes to call this option record `GAPDocLaTeXOptions` or something like this, so it would make not so much sense to convert it to snake case. If anything, lets just go with camel for making it more gappy.

gap/Magic.gd
sebasguts: I guess you mean the `file` option here? Guess it is really old, and yeah, you are probably right it can be done better. So if it comes to me, we can just remove it.

sebasguts: And yeah, it seems like the code was added in earlier times by me. It uses `StringFile` though, so that is something ;)

Issue #205, opened 2019-03-28
Deprecate @BeginSystem, @EndSystem, and @InsertSystem


fingolfin:
I could not find any packages using `@BeginSystem` or `@EndSystem`, so I think we could just remove those two right away.

For `@InsertSystem`, I found several (meta) packages using it:
- [x] ArangoDBInterface
- [x] ToricVarieties
- [x] GroupRepresentationsForCAP
- [x] homalg_project_extensions/Blocks
- [x] homalg_project_extensions/SingularForHomalg
- [x] homalg_project_extensions/PrimaryDecomposition
- [x] homalg_project_extensions/LessGenerators

So, all of these are from Siegen, so I am hopefully we can get them to accept PRs that change `@InsertSystem` to `@InsertChunk`. Once that's done, and there were releases, we can remove `@InsertSystem`, too.

sebasguts:
Will provide some PR's

fingolfin:
Thanks. It'd be great if you could edit the text above and insert links to those PRs, that'll help us track the status of this.

fingolfin:
I've now provided those PR instead, and most of them have been merged, only waiting for https://github.com/homalg-project/SingularForHomalg/pull/2 but I am hopeful @mohamed-barakat will merge it soon. I guess we then still should wait a bit with releasing a new version that removes these chunks... In the meantime, I'll prepare a PR to print out a warning when they are used.

mohamed-barakat:
@fingolfin: Merged, thanks.

sebasguts:
Ah sorry, totally forgot about this.

Are we sufficiently sure nobody else uses those commands? Well, I guess we cannot take care of code that is not available somewhere.

fingolfin:
Resolved by deprecating (instead of removing) these commands; deprecation here means that a warning is printed in red.
We can perform the actual removal at some day in the future (say with GAP 4.12 or 4.13), or simply leave these in for ever (but perhaps remove them from the manual at some point).

Issue #206, opened 2019-03-29
Deprecate @EndChapter, @EndSection, @EndSubsection


fingolfin:
Deprecate (and eventually, remove) `@EndChapter`, `@EndSection`, and `@EndSubsection`. Instead, improve the logic which automatically ends chapters/sections/subsections. This is motivated by #157 and @oysteins comments there. To steal from his comment there:

- [ ] Make subsections end automatically when encountering a declaration.
- [ ] Then deprecate the `@EndChapter`, `@EndSection`, and `@EndSubsection` commands in the manual; also change them print warnings (?); and provide PRs to the handful of packages using them:
  - [ ] datastructures
  - [ ] LocalNR
  - [ ] HeLP
- [ ] once all packages using them have been updated (resp. had a release?) we can fully remove them (or set a date for removal, adding a note in the manual and/or warnings saying something like "these will be removed in 2020")
- [ ] Explain in the manual that a subsection and a declaration give headings on the same level (or in other words: a declaration gives a special type of subsection), which means that one cannot use subsections to structure the declarations.
- [ ] optionally: consider supporting `# Chapter`, `## Section` and `### Subection` syntax, at least in `.autodoc` files...

sebasguts:
I would not like to deprecate them all. `@EndChapter` might be somehow useless, although it would prevent from putting further method declaration ending in that chapter, but in the auto generated chapter.

However, `@EndSection` and `@EndSubsection` can be used to structurize text. When you do a subsection and end it, text inputed afterwards will be in the section, not the subsection. I am not sure if this is of any immediate use, but it reflects the way the HTML is structured.

fingolfin:
OK, I guess I somewhat see a point for `@EndSubsection` then. But for `@EndSection`, I don't -- unless you want to allow for text *between* sections, which seems like a bad idea, as explained in issue #157.

And I'd argue that even text between subsections is a bad idea (there is a reason why neither MarkDown nor LaTeX support that, I guess). And it is usually impossible for the reader to tell the difference visually anyway. And what would one use this for? What kind of text would not instead also fit into the end of one subsection or the start of the next?

Of course, I am in general sceptical about using subsections *at all*, so I might be overly negative :-)

sebasguts:
I see. Well, putting the introductory text at the beginning of a chapter is one of my main reasons.

As you correctly say, latex does not really support this feature anyway. We could (first silently) deprecate them, so we can just make them stop do anything at all. Then drop them at some point.

[...]

PR #212, opened 2019-05-17
Implicitly end subsection when a manpage starts; add deprecation warnings

fingolfin:
Specifically, add deprecation warnings for @InsertSystem, @System, @BeginSystem, @EndSystem (use @Chunk etc. instead), and also @EndSection, @EndSubsection.

These warnings are shown for info level InfoWarning >= 1, and in red color to make sure they are seen by the user.

Also, the manual was changed to document these tags as deprecated, and what to use instead.

This mostly resolves the issues #205 and #206; however, we might keep those open to remind us that one day in the future (let's say, in one year?) we could make a release were those deprecated tags are completely removed; or at least removed from the manual.

### DIFF
--- a/gap/Parser.gi
+++ b/gap/Parser.gi
@@ -135,7 +135,7 @@ InstallGlobalFunction( AutoDoc_Parser_ReadFiles,
           level_scope, scope_group, read_example, command_function_record, autodoc_read_line,
           current_command, was_declaration, filename, system_scope, groupnumber, chunk_list, rest_of_file_skipped,
           context_stack, new_man_item, add_man_item, Reset, read_code, title_item, title_item_list, plain_text_mode,
-          current_line_unedited,
+          current_line_unedited, deprecated,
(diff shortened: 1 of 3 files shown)


gap/Parser.gi
```diff
@@ -168,10 +168,18 @@ InstallGlobalFunction( AutoDoc_Parser_ReadFiles,
...
+        if IsBound( chapter_info[ 3 ] ) then
+            Unbind( chapter_info[ 3 ] );
+            current_item := SectionInTree( tree, chapter_info[ 1 ], chapter_info[ 2 ] );
+        fi;
+
         if IsBound( current_item ) then
             Add( context_stack, current_item );
         fi;
+
+        man_item := DocumentationManItem( tree );
```
sebasguts: I guess you made sure that I can push the current item on the context stack before creating the new `man_item`. It makes much more sense this way anyway . . .

PR #234, opened 2020-01-13
Add support for DeclareFamilyProperty

zickgraf:
This is a CAP-specific version of DeclareProperty with the signature
`DeclareFamilyProperty( name, filter[, family, type of cell ] )`

The CAP-PR making use of this is https://github.com/homalg-project/CAP_project/pull/459

@sebastianpos are you aware of further CAP-specific versions of declarations?

### DIFF
--- a/PackageInfo.g
+++ b/PackageInfo.g
@@ -10,7 +10,7 @@ SetPackageInfo( rec(
 
 PackageName := "AutoDoc",
 Subtitle := "Generate documentation from GAP source code",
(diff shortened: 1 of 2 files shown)


zickgraf:
@sebastianpos just mentioned [here](https://github.com/homalg-project/CAP_project/pull/459#issuecomment-573748651) that CAP could simply get rid of `DeclareFamilyProperty`. So I will close this.

PR #245, opened 2020-08-07
Remove outdated extracted tst files

### LINKED ISSUE #240: problem with the option extract_examples
The function `AUTODOC_ExtractMyManualExamples` should first remove all testfiles whose name follows the scheme of the files it creates.
Currently it can happen that outdated testfiles remain --this happens when the number of those chapters decreases for which manual examples exist.

fingolfin:
Resolves #240

### DIFF
--- a/gap/AutoDocMainFunction.gi
+++ b/gap/AutoDocMainFunction.gi
@@ -423,15 +423,36 @@ end );
 BindGlobal("AUTODOC_ExtractMyManualExamples",
 function( pkgname, pkgdir, docdir, main, files, opt )
     local tst, i, s, basename, name, output, ch, a, location, pos, comment, pkgdirString,
-      nonempty_units_found, number_of_digits;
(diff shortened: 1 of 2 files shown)


gap/AutoDocMainFunction.gi
```diff
@@ -456,9 +474,8 @@ function( pkgname, pkgdir, docdir, main, files, opt )
         fi;
         # pad s to number_of_digits
         s := Concatenation( ListWithIdenticalEntries( number_of_digits - Length( s ), '0' ), s );
-        basename := Concatenation( LowercaseString(pkgname), s, ".tst" );
-        basename := ReplacedString( basename, " ", "_" );
```
fingolfin, line 460: I think this was not doing anything, not sure why it was even there (package names are not allowed to contain spaces)

fingolfin: Ah I was wrong, this can be almost anything for worksheets. But then the old code was kinda wrong, too. Will fix it.

PR #277, opened 2024-03-25
change referenced file

cdwensley:
Changed makedoc.g to PackageInfo.g, where the names of io and PackageName are defined.
Found this error where searching how to define/use entity &lt which is not working for me.

### DIFF
--- a/doc/Tutorials.xml
+++ b/doc/Tutorials.xml
@@ -36,7 +36,7 @@ LoadPackage( "AutoDoc" );
 AutoDoc( rec( scaffold := true ) );
 </Listing>
 This first reads the <F>PackageInfo.g</F> file from the current
-directory. It extracts information about package from it
+directory. It extracts information about the package from it
 (such as its name and version, see Section <Ref Sect="Tut:Scaffolds:Title"/>).
 It then creates two XML files <F>doc/NAME_OF_YOUR_PACKAGE.xml</F> and
 <F>doc/title.xml</F> inside the package directory. Finally, it runs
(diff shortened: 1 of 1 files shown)


doc/Tutorials.xml
```diff
@@ -383,7 +383,7 @@ Edit <File>PackageName/makedoc.g</File> as follows.
   In the <Code>entities</Code> record enter any entities 
   previously stored in <File>manual.xml</File>. 
   (Again, if you have none, you may safely delete this record.) 
-  To illustrate this option the &AutoDoc; file <File>makedoc.g</File> 
+  To illustrate this option the &AutoDoc; file <File>PackageInfo.g</File>
```
fingolfin, line 386: Thank you for your patch, which is much appreciated. Unfortunately it is not quite ready to be merged as-is:

Just above the fold you can see `<Index Key="Entities record in makedoc.g"></Index>`, so this block of text really is about `makedoc.g`.

At some point I've added the ability to *also* specify entities in `PackageInfo.g`, and carelessly removed the code for doing so from `makedoc.g` in ace699ca7ac02cd1980192e6babd84d295212d56 as I had forgotten about this mention in the manual... There's a limit how many techniques one can demonstrate in one package ;-).


The thing is, all these records can be specified in either `PackageInfo.g` or `makedoc.g`, and which is preferred depends on the package authors. I guess the text should explain these choices, and perhaps also show code snippets inline... I can try to look into doing that when I find time -- or, well, of course a PR by someone else would also be fine :-).

So I won't merge this PR as-is, but there definitely is something to be fixed and improved here. So thank you very much for drawing my attention to this.

cdwensley:
Working on it.

cdwensley:
The last item under 'editing makedoc.g' has been moved to 'editing PackageInfo.g'.
Several minor corrections made earlier.
Commented out the sentence "For other introductory components, such as Colophon, consult the file gap/Magic.gd."
That's because Magic.gd has nothing about it, and nor does any other file.

PR #280, opened 2024-11-12
let `AutoDoc` set an exit code in the case of failure

ThomasBreuer:
The idea is the same as in gap-system/gap/pull/5835, but here we deal with package documentation:
When building a manual, collect the info messages printed by `MakeGAPDocDoc`, use a heuristic to decide whether the result is acceptable, and let `makedoc.g` report failure if not; this can be used in automated tests.

For that, the function `AutoDoc` now ~~returns a record; up to now, it had no return value.~~ sets a nonzero exit code if applicable.

This change will automatically become available in all packages that use `makedoc.g` containing `AutoDoc` calls. This should not cause problems because the exit code of processing `makedoc.g` did not have a meaning before the change.

### DIFF
--- a/gap/Magic.gd
+++ b/gap/Magic.gd
@@ -265,6 +265,11 @@
 #!         In all other cases (in particular if <A>opt.gapdoc</A> is
 #!         <K>false</K>), this feature is disabled.
 #!         <P/>
+#!         If &GAPDoc; is invoked by <Ref Func="AutoDoc"/> then the returned
+#!         record has a component <C>GAPDoc_Info</C>, whose value is the
+#!         string obtained from collecting the info messages printed by
+#!         the &GAPDoc; functions.
+#!         <P/>
 #!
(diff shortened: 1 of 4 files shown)


ThomasBreuer:
@fingolfin
From the viewpoint of https://github.com/gap-system/ReleaseTools/blob/master/release-gap-package, it is not a good idea to change `AutoDoc` to return something:
The `makedoc.g` files of packages that call `AutoDoc` typically do not suppress its output, and the `release-gap-package` script collects and evaluates all output obtained by processing `makedoc.g`.
Thus the proposed change would make `release-gap-package`'s work more complicated, instead of simplifying the task.
I will change the approach (again) ...

fingolfin:
We are now catching this kind of issue directly in the relevant GitHub Actions, which scan for these outputs and more. So we don't need this approach anymore.

PR #283, opened 2025-03-26
Add InfoAutoDoc for messages instead of Printing

mtorpey:
In various places, AutoDoc uses `Print` to print status messages, with no way to silence these.  This is a bit annoying in automated workflows like the package manager.

This PR adds a new info class `InfoAutoDoc` which is used instead of `Print`.  It's currently binary: 1 for messages and 0 for no messages.  The default is 1, of course.

### DIFF
--- a/gap/Magic.gi
+++ b/gap/Magic.gi
@@ -215,7 +215,7 @@ function( arg )
     if IsBound(scaffold) and IsBound( pkginfo.AutoDoc ) then
         for key in RecNames( pkginfo.AutoDoc ) do
             if IsBound( scaffold.(key) ) then
-                Print("WARNING: ", key, " specified in both PackageInfo.AutoDoc and opt.scaffold\n");
+                Info(InfoGAPDoc, 1, key, " specified in both PackageInfo.AutoDoc and opt.scaffold");
             else
                 scaffold.(key) := pkginfo.AutoDoc.(key);
             fi;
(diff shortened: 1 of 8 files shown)


mtorpey:
CodeCov is the only failing test here – all lines I've added are tested, it just seems to want to test some lines I've changed, which I'm not exactly sure how to do.

From what I can tell, this is okay.

gap/Magic.gi
```diff
@@ -215,7 +215,7 @@ function( arg )
     if IsBound(scaffold) and IsBound( pkginfo.AutoDoc ) then
         for key in RecNames( pkginfo.AutoDoc ) do
             if IsBound( scaffold.(key) ) then
-                Print("WARNING: ", key, " specified in both PackageInfo.AutoDoc and opt.scaffold\n");
+                Info(InfoGAPDoc, 1, key, " specified in both PackageInfo.AutoDoc and opt.scaffold");
```
fingolfin, line 218: This should retain the word "Warning"

mtorpey:
I've made the changes suggested above.

tst/worksheets.tst
```diff
@@ -5,16 +5,9 @@ gap> START_TEST( "worksheets.tst" );
 
 #
 gap> AUTODOC_TestWorkSheet("general");
-Extracting manual examples for General Test package ...
-2 chapters detected
-Chapter 1 : extracted 2 examples
-Chapter 2 : no examples
```
fingolfin, line 11: Any particular reason to turn off the info messages in the test suite here? I actually found it helpful to see this for debugging purposes...

mtorpey: No reason in particular!

Normally I turn them off in test suites because they contain things like paths or random elements that are difficult to reproduce. But in this case I don't think that's a problem.

I'll turn them back on, and also fix something else I just spotted…

mtorpey: They're back on in the test suite, which actually revealed a mistake I made, which I've now fixed. So they've already been useful!

mtorpey:
I've made a few more changes. `InfoAutoDoc` is now on in the test suite, while InfoGAPDoc is off.  I also introduced level 2 for `InfoAutoDoc` which includes the "Generating documentation in <directory>" messages, which are useful but system-dependent so are turned off in tests.

I'm happy with this now if you want to take another look.

fingolfin (approved):
Thank you!!!

Issue #284, opened 2025-06-05
Too long homepage URLs are cut off in the PDF file


Sebas777-gif:
This is a rather banal problem, but specifying a string for `WWWHome` for a person in `PackageInfo.g` that is 96 characters long results in the URL being cut off in the PDF version of the documentation. The link is still clickable, so this problem is of purely aesthetic nature.

If there is a workaround for this that I have not thought of, I apologize; I am rather new to GAP development.

fingolfin:
Hi @Sebas777-gif thank you for your report. Unfortunately I didn't see your issue and PR before today (I get dozens and sometimes hundreds of notifications from GitHub per day and can't check them all -- so it usually is best to also send me an email about issues. Or have patience... :-)  )

I am not sure this is the best way to address this. Instead, I think this should be addressed in GAPDoc (ping @frankluebeck). E.g. by enabling line-wrapping for that URL there; or by just not printing the URL.

PR #285, opened 2025-06-05
Makeshift solution for too long URLs not being cut off in the PDF

Sebas777-gif:
Addresses #284.
This is a _very_ ugly workaround and also has the problem that the homepage URLs are now not in monospace font anymore. I did not find a way to change the font type.
I'd be very glad for suggestions of better approaches to this problem; I am still a GAP novice. Thank you.

### DIFF
--- a/gap/AutoDocMainFunction.gi
+++ b/gap/AutoDocMainFunction.gi
@@ -168,7 +168,7 @@ end );
 ##
 InstallGlobalFunction( ExtractTitleInfoFromPackageInfo,
   function( pkginfo )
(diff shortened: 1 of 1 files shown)


frankluebeck:
I just found the `ModularGroup` repository, and the manual looks fine to me.

Can this be closed?

In general it is very complicated to make LaTeX hyphenate text in typewriter font.

fingolfin:
@frankluebeck I've closed this on March 6. Did you perhaps mean to comment on another issue? For ModularGroup, we put in a workaround some time ago.

PR #286, opened 2025-09-04
Delete some long-deprecated code

fingolfin:
... to simplify the code, enabling further refactoring later on.


fingolfin:
CC @mohamed-barakat not sure if you guys still use `maketest` anywhere (it seems to be in some homalg-project repositories, but mostly only old ones?

PR #322, opened 2026-03-06
Remove broken level commands

fingolfin:

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,11 @@
 This file describes changes in the AutoDoc package.
 
 2026.MM.DD
+  - Remove the nonfunctional `@Level`, `@ResetLevel`, and undocumented
+    alias `@SetLevel` commands. They never affected generated output as
+    documented. Since nobody ever reported issues with them, and since
+    no distributed packages uses them, they probably are simply not in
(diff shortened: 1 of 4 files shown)


fingolfin:
Thanks @mohamed-barakat . I started by doing `gap makedoc.g` in that repository and the first error *I* did run into actually was about `@URL` not being a valid command. I had a look: we removed that in release 2016.02.24 😂 

So I removed that line, and now it indeed complained about an invalid date:
```
Error, Invalid arguments at /Users/mhorn/Projekte/GAP/repos/pkg/gap-packages/AutoDoc/gap/ToolFunctions.gi:290 called from
AUTODOC_FormatDate( argument_rec.Date
```
Indeed it tries to parse the date as if it was a date in a `PackageInfo.g` -- but that's a bug. Also, the error is bad, it gives no indication where it comes from (so another issue to fix). But least it suggest a temporary workaround: replace e.g. `@Date 16. February 2015` by `@Date 2015-02-16`. Alas, that *still* fails because it actually extracts the string `"2015-02-16\n"`, so with a trailing newline -- and that chokes it *sigh*.

I'll work on a fix (or rather, will let Codex make one)

Issue #78, opened 2014-09-11
Improve code translating chapter / section titles to filenames


fingolfin:
The documentation for @Chapter and @Section should state explicitly that : and \ are not allowed in them.

Issue #77, opened 2014-09-11
Better error message when chapter or section name contains invalid character


fingolfin:
I wrote some documentation and dared to put a colon into a chapter title. This triggered the following error:
```
Generating documentation in Directory("/PATH/TO/doc/")
Error, <name> must not contain '\' or ':' called from
Filename( arg[1], arg[2] ) called from
AUTODOC_OutputTextFile( path_to_xmlfiles, filename ) called from
WriteDocumentation( i, stream, path_to_xmlfiles ); called from
WriteDocumentation( tree, doc_dir ); called from
<function "AutoDoc">( <arguments> )
 called from read-eval loop at line 21 of makedoc.g
you can 'quit;' to quit to outer loop, or
you can 'return;' to continue
brk>
```
That is not very helpful. I had written a big chunk of stuff, and part of it contained a backslash. After removing that, I still got errors, and it took me some time that I should search for a colon...

I think we should check chapter and section names the moment we read them in, and *immediately* print an error at that point, if necessary -- including filename, line and the precise string that was read...

fingolfin:
Note that with my changes in a4de35771fc331437fa4ec83946a035de4f99cdb it should now be relatively (?) easy to implement a better error message. Basically, we need a function which validates a given any chapter / section / subsection title we just read. And if the title contains characters outside a certain predefined set, the function raises an error, pointing out the file and line with the "bad" title.

fingolfin:
Alternatively, we could simply remove this error, by following the discussion in #78

PR #333, opened 2026-03-08
Ignore trailing blank lines after single-line title-page commands.

fingolfin:
Trim generated title-page content to avoid stray newlines.
This prevents newline-bearing filenames and empty authors.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,10 @@ This file describes changes in the AutoDoc package.
     `<Keyword>...</Keyword>` for GAP keywords (as returned by
     `ALL_KEYWORDS()`), otherwise as before `<Code>...</Code>`
   - Allow XML-style comments in `.autodoc` files
+  - Ignore trailing blank lines after single-line worksheet title-page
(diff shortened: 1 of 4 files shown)


fingolfin:
@mohamed-barakat with that the worksheet in your private repo builds again (well, with the bib file fix (needed by GAPDoc) and the removal of that `@URL` command (which we dropped support for over 10 years ago; but we may soon have Markdown style links as a better replacement). Afterwards, there are still some broken Refs: apparently if a chapter or (sub)section title contains characters like `:`, then those are stripped from auto-generated labels. I don't know yet if that's a recent regression or a change we made long ago (depending on that I can revert it, or not...).  That said, I am not super happy with our current auto-label system, but that's another matter.

fingolfin:
@mohamed-barakat ah the changed chapter labels is likely due to the fixes for #77 and #78, in commit 2ec2bed. I guess that change was overly restrictive, and we could either add more character to the "allow list", or just block certain characters (say: any control characters, slashes, backslashes). I'll look into it.

PR #363, opened 2026-03-18
Remove or deprecate various legacy features

fingolfin:
- Remove the undocumented AutoDoc package-name invocation.
- Stop honoring GAP global options as overrides for normal AutoDoc options.
- Keep worksheet compatibility, with a warning for legacy ':' syntax.

The main motivation of this is to reduce complexity, and remove features that are never tested by us (e.g. because they are hard to test). In fact for most of these, Sebastian agreed many many years ago that we don't want them anymore. I am not aware of any users out there "in the wild".

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,19 @@
 This file describes changes in the AutoDoc package.
 
+## 2026.MM.DD
+
++ **Breaking changes**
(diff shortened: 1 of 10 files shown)


fingolfin:
I've found several packages still using this. So instead of a hard removal, it might be better to first show deprecation warnings for a while. Then again, is anyone going to actually read and heed those? Hm

PR #369, opened 2026-05-09
Look for GAP/autodoc files in pkgdir first, and in the current directory second

stertooy:
This prevents problems in the rare case where we are calling a `makedoc.g` script from outside the package's directory, but the current directory has a file with the same *relative* path as one of the files used in the documentation.

For example, this can happen by calling `makedoc ../packageA/makedoc.g` from within Package B's directory (this is actually how I encountered the problem). And that's not an entirely unreasonable thing to do, since package B's documentation may require Package A's documentation to be compiled first, in order to resolve references.

### DIFF
--- a/gap/Magic.gi
+++ b/gap/Magic.gi
@@ -228,11 +228,11 @@ function( is_worksheet, pkgname, pkginfo, pkgdir, opt )
         # Make sure all of the files exist, making the file names absolute if
         # necessary
         for i in [ 1 .. Length( autodoc.files ) ] do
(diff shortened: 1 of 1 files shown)


fingolfin:
Good catch, thanks

You probably are aware, but just to make sure: note that until very recently, this was not supported at all: we always searched in the current directory. So if you invoked `gap ../packageA/makedoc.g` from within `packageB`, it built the documentation of `packageB`, not of `packageA`.

fingolfin (approved):

PR #372, opened 2026-07-01
WriteChunks: write all chunks before checking insertion status

ChrisJefferson:
Writing a chunk whose body contains a nested @InsertChunk marks the inserted chunk as inserted. Previously the "defined but never inserted" and "inserted but never defined" checks were interleaved with writing, so a chunk that was only inserted from within a later-written chunk could be spuriously reported, depending on iteration order. Split the loop: write every chunk first, then run the checks once all insertions have been resolved.

Found while fixing the vole package.

AI disclosure: this change was prepared with the assistance of Claude Code (Anthropic), which diagnosed the ordering issue and drafted the fix.


fingolfin:
Thanks @ChrisJefferson . My Claude added a test and a changelog entry.

